PR #21, opened 2021-11-29
Markdown Notebooks

zampino:
- Separate top-level clojure forms grouped under the same markdown cell
- Handle markdown indented code blocks as literal code
- Parse markdown chunks in clojure inputs as well unifying notebook
representations for markdown and clojure files.

### DIFF
--- a/src/nextjournal/clerk/hashing.clj
+++ b/src/nextjournal/clerk/hashing.clj
@@ -122,15 +122,17 @@
 
 #_(read-string "(ns rule-30 (:require [nextjournal.clerk.viewer :as v]))")
 
-(def tag-allowed? #{:deref :map :meta :list :quote :reader-macro :set :token :var :vector})
+(def code-tags
+  #{:deref :map :meta :list :quote :reader-macro :set :token :var :vector})
+
 (defn parse-clojure-file
   ([file] (parse-clojure-file {} file))
   ([{:as _opts :keys [markdown?]} file]
-   (loop [{:as state :keys [doc nodes visibility]} {:nodes (:children (p/parse-file-all file))
-                                                    :doc []}]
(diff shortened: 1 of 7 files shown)


zampino:
what should be the convention for markdown code blocks to be evaled? Currently:

- 2-stops indented code is rendered markdown inert code
- fenced code blocks are evaluated

what about fence-info with an explicit language? If it's set then only eval if it equals `clojure` or `clj` ?

src/nextjournal/clerk/hashing.clj
```diff
@@ -140,14 +142,58 @@
...
+(defn code-cell? [{:as node :keys [type]}]
+  ;; TODO: assign different types to fenced vs indented code blocks
+  (and (= :code type) (contains? node :info)))
+
+(defn parse-markdown-file [{:keys [markdown?]} file]
+  (loop [{:as state :keys [nodes] ::keys [md-slice]} {:doc [] ::md-slice [] :nodes (:content (markdown/parse (slurp file)))}]
+    (if-some [node (first nodes)]
+      (recur
+       (if (code-cell? node)
+         (let [apply-clj-nodes (fn [state clj-nodes]
```
zampino, line 161: this _local_ var instantiates a function at every node, this is bad, right?

src/nextjournal/clerk/view.clj
```diff
@@ -119,9 +119,9 @@
   ([{:keys [inline-results?] :or {inline-results? false}} doc]
    (let [{:keys [ns]} (meta doc)]
      (cond-> (into []
-                   (mapcat (fn [{:as cell :keys [type text result ns?]}]
```
zampino, line 122: `ns?` was unused.

PR #24, opened 2021-12-08
replace function `hide-result` with metadata

behrica:


mk:
Hi, thanks for your PR. We thought about using metadata like in this PR to control result hiding but ended up favoring the current implementation of using a viewer to control visibility so I'm not going to merge this for now.

This is not to say that the current implementation is final, we've actually been talking about using metadata for setting the viewers more so there's a good chance things will still change in the future.

mk:
Using `clerk/hide-result` more we're coming around that this is the better (less annoying) approach.

mk:
There's still an issue here that the result isn't shown when e.g. setting `:fold` visibility on the ns as in notebooks/visibility.clj.

PR #26, opened 2021-12-09
Add first tests & refactor analysis

philomates:
I was spending some time understanding what is happening in the hash analysis logic and I usually understand code by doing refactor passes on it. Figured I'd post them as a PR in case y'all find them helpful; I'm happy to revert some or all changes that don't agree with folks.

I ended up converting some of the repl-able inline comments into tests because my initial refactor pass broke some stuff.


src/nextjournal/clerk/hashing.clj
```diff
@@ -22,13 +22,15 @@
 (defn var-name
   "Takes a `form` and returns the name of the var, if it exists."
   [form]
-  (when (and (sequential? form)
+  (when (and (list? form)
```
philomates, line 25: vectors are sequentials but you can't have `def`/`defn`s using vectors, so switching to `list?` is a bit tighter; added https://github.com/nextjournal/clerk/pull/26/files#diff-b33fbf28a0d44915162c46241d20d137f2ca22398ab87a4881637a3358159cc8R22 to check this

src/nextjournal/clerk/hashing.clj
```diff
@@ -199,40 +200,53 @@
...
+                            (dep/depend var rec-var)
+                            (dep/depend dep rec-var)))
+        (assoc-in [:var->hash rec-var :form] rec-form))))
+
+(defn- analyze-deps [var form {:as state :keys [graph]} dep]
+  (try (assoc state :graph (dep/depend graph (if var var form) dep))
+       (catch Exception e
+         (when-not (circular-dependency-error? e)
+           (throw e))
+         (analyze-circular-dependency state var form dep (ex-data e)))))
```
philomates, line 221: none of the in-line repl examples hit this code path so these changes aren't verified :grimacing: maybe we can extend the test coverage before going forward with this change

src/nextjournal/clerk/hashing.clj
```diff
@@ -199,40 +200,53 @@
...
+
+(defn- analyze-codeblock [file state {:keys [type text]}]
+  (let [{:keys [var deps form ns-effect?]} (-> text read-string analyze)
+        state-with-var-hash (assoc-in state
+                                      [:var->hash (if var var form)]
+                                      {:file file
+                                       :form form
+                                       :deps deps})]
+    (when ns-effect?
+      (eval form))
```
philomates, line 231: I don't fully understand the clerk workflow but in trying to use `clerk/show!` I found it was changing the ns of my repl. Would it be useful to capture and restore the current ns before running stuff ns-effectful expressions? Maybe similar to https://github.com/nextjournal/clerk/pull/26/files#diff-b33fbf28a0d44915162c46241d20d137f2ca22398ab87a4881637a3358159cc8R8-R12

src/nextjournal/clerk/hashing.clj
borkdude: I would not ever use `list?` but use `seq?` instead. `list?` will return false for "lists" created by syntax-quote expressions which are typically constructed using `cons`.

Issue #27, opened 2021-12-10
Watcher does not refresh file on change event on OS X


ikappaki:
Hi, 

there appears to be a recent issue on OS X that watcher fails to load file after it has been modified.

to reproduce

1. Bring up the REPL and watch the `notebooks` folder
```clojure
(require '[nextjournal.clerk :as clerk])
(clerk/serve! {:watch-paths ["notebooks"]})
```
2. Open `http://localhost:7777`. and back at the REPL do
```clojure
(clerk/show! "notebooks/hello.clj")
```
3. The `hello.clj` should be shown at the browser.
4.  Add something to the `hello.clj` file and save it.
4. The page should refresh with the addition, but nothing happens.

PR to follow.

mk:
Closed with 8ad88630f746f1a9ff3ac314f5528c2d25c42583.

PR #23, opened 2021-12-04
Ignore Emacs lock files from visualising

ikappaki:
Hi,

could you please consider patch to ignore files whose names begins with `.#` because are most likely Emacs lock files that `Clerk` might try to visualize on MS-Windows as described in #22?

I've introduce the first unit test to test the same (happy to remove if you think is too much or you have other plans for unit testing).

Lastly, I've changed `git:/` to `https:/` in deps, since these are likely to cause unexpected blocking issues on MS-Windows (see https://github.com/nextjournal/viewers/pull/30 for a description).

I've also added Emacs temp files ending with `~` in`.gitignore`. 

Also upgraded `shadow-cljs` version to 2.16.7 to solve issue with gitlibs when they are stored on a separate drive than the project (see https://github.com/thheller/shadow-cljs/issues/966 for the details).

Please feel free to scrutinize.

Thanks.

Closes #22.

PR #28, opened 2021-12-10
Fix file watcher regression

ikappaki:
Hi,

could you please consider fix for issue #27.

It is caused by the recent #23 commit which uses a regexp against the modified filename to determine which files are supported to be shown by the browser.

The original commit converted the modified file's `Path` into an `io/file` for convenience and then extracted the filename as a `String`, but the `Path` to `io/file` conversion fails on OS X throwing an exception (as shown at the end).

The fix is to retrieve the filename string from the `Path` directly.

Also, it might be a good idea to have `clerk/file-event` `try` and `catch` any exception thrown instead of silently failing?

error thrown: (is this a bug?)
```clojure
:error #error {
 :cause No implementation of method: :as-file of protocol: #'clojure.java.io/Coercions found for class: sun.nio.fs.UnixPath
 :via
 [{:type java.lang.IllegalArgumentException
   :message No implementation of method: :as-file of protocol: #'clojure.java.io/Coercions found for class: sun.nio.fs.UnixPath
   :at [clojure.core$_cache_protocol_fn invokeStatic core_deftype.clj 583]}]
 :trace
 [[clojure.core$_cache_protocol_fn invokeStatic core_deftype.clj 583]
  [clojure.java.io$fn__11390$G__11372__11395 invoke io.clj 35]
  [clojure.java.io$file invokeStatic io.clj 424]
  [clojure.java.io$file invoke io.clj 418]
  [nextjournal.clerk$supported_file_QMARK_ invokeStatic NO_SOURCE_FILE 181]
  [nextjournal.clerk$supported_file_QMARK_ invoke NO_SOURCE_FILE 174]
  [nextjournal.clerk$file_event invokeStatic clerk.clj 211]
  [nextjournal.clerk$file_event invoke clerk.clj 209]
  [nextjournal.clerk$serve_BANG_$fn__18760 invoke clerk.clj 289]
  [nextjournal.beholder$fn$reify__6032 onEvent beholder.clj 13]
  [io.methvin.watcher.DirectoryWatcher onEvent DirectoryWatcher.java 352]
  [io.methvin.watcher.DirectoryWatcher watch DirectoryWatcher.java 297]
  [io.methvin.watcher.DirectoryWatcher lambda$watchAsync$0 DirectoryWatcher.java 212]
  [java.util.concurrent.CompletableFuture$AsyncSupply run CompletableFuture.java 1768]
  [java.util.concurrent.CompletableFuture$AsyncSupply exec CompletableFuture.java 1760]
  [java.util.concurrent.ForkJoinTask doExec ForkJoinTask.java 373]
  [java.util.concurrent.ForkJoinPool$WorkQueue topLevelExec ForkJoinPool.java 1182]
  [java.util.concurrent.ForkJoinPool scan ForkJoinPool.java 1655]
  [java.util.concurrent.ForkJoinPool runWorker ForkJoinPool.java 1622]
  [java.util.concurrent.ForkJoinWorkerThread run ForkJoinWorkerThread.java 165]]}
```

(Tested it to work on both OS X and MS Windows)

Thanks!

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -175,16 +175,14 @@
   "Returns whether `path` points to a file that should be shown."
   [path]
   ;; file names starting with .# are most likely Emacs lock files and should be ignored.
-  (->> path .getFileName .toString
-       (re-matches #"(?!^\.#).+\.(md|clj|cljc)$")
-       some?))
+  (some? (re-matches #"(?!^\.#).+\.(md|clj|cljc)$" (.. path getFileName toString))))
 
-#_(supported-file? (-> "foo_bar.clj" (java.nio.file.Paths/get (into-array String []))))
-#_(supported-file? (-> "xyz/foo.md" (java.nio.file.Paths/get (into-array String []))))
-#_(supported-file? (-> "xyz/foo.clj" (java.nio.file.Paths/get (into-array String []))))
-#_(supported-file? (-> "xyz/a.#name.cljc" (java.nio.file.Paths/get (into-array String []))))
-#_(supported-file? (-> ".#name.clj" (java.nio.file.Paths/get (into-array String []))))
-#_(supported-file? (-> "xyz/.#name.cljc" (java.nio.file.Paths/get (into-array String []))))
+#_(supported-file? (fs/path "foo_bar.clj"))
+#_(supported-file? (fs/path "xyz/foo.md"))
+#_(supported-file? (fs/path "xyz/foo.clj"))
+#_(supported-file? (fs/path "xyz/a.#name.cljc"))
+#_(supported-file? (fs/path ".#name.clj"))
+#_(supported-file? (fs/path "xyz/.#name.cljc"))
 
 
 (defn file-event [{:keys [type path]}]
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -175,16 +175,16 @@
   "Returns whether `path` points to a file that should be shown."
   [path]
   ;; file names starting with .# are most likely Emacs lock files and should be ignored.
-  (->> path io/file .getName
+  (->> path .getFileName .toString
        (re-matches #"(?!^\.#).+\.(md|clj|cljc)$")
        some?))
 
-#_(supported-file? "foo_bar.clj")
-#_(supported-file? "xyz/foo.md")
-#_(supported-file? "xyz/foo.clj")
-#_(supported-file? "xyz/abc.#name.cljc")
-#_(supported-file? ".#name.clj")
-#_(supported-file? "xyz/.#name.cljc")
+#_(supported-file? (-> "foo_bar.clj" (java.nio.file.Paths/get (into-array String []))))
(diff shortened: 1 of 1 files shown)


ikappaki:
It appears `clojure.java.io/file` does not support `Path`s as an argument to begin with, so original implementation was incorrect trying to coerce `Path` into a `File`. It should fail on all architectures, not just OS X.

https://clojure.atlassian.net/browse/CLJ-2333 

`io/file` delegates to the `io/Coercions` protocol's `as-file` method for coercing its arguments into a `File`. It can [coerce `String`, `File`, `URL` and `URI`s](https://github.com/clojure/clojure/blob/abe19832c0294fec4c9c55430c9262c4b6d2f8b1/src/clj/clojure/java/io.clj#L44)  by default.

It so happens on my main dev machine I have [clj-refactor](https://github.com/clojure-emacs/clj-refactor.el) installed, which injects the [refactor-nrepl](https://github.com/clojure-emacs/refactor-nrepl) into the repl, which inlines [clj-commans/fs](https://github.com/clj-commons/fs) (with [MrAnderson](https://github.com/benedekfazekas/mranderson)) , which extends [`io/Coercions` with Path](https://github.com/clj-commons/fs/blob/96b73d49f163f17b22b93bfd35fd9322f2785c10/src/me/raynes/fs.clj#L144), and thus how it was tested to work for me in the first place ..., e.g. in my REPL with clj-refactor installed , I get:

```clojure
user> (:impls clojure.java.io/Coercions)
{nil
 {:as-file #function[clojure.java.io/fn--11402],
  :as-url #function[clojure.java.io/fn--11404]},
 java.lang.String
 {:as-file #function[clojure.java.io/fn--11406],
  :as-url #function[clojure.java.io/fn--11408]},
 java.io.File
 {:as-file #function[clojure.java.io/fn--11410],
  :as-url #function[clojure.java.io/fn--11412]},
 java.net.URL
 {:as-url #function[clojure.java.io/fn--11414],
  :as-file #function[clojure.java.io/fn--11416]},
 java.net.URI
 {:as-url #function[clojure.java.io/fn--11418],
  :as-file #function[clojure.java.io/fn--11420]},
 java.nio.file.Path
 {:as-file
  #function[refactor-nrepl.inlined-deps.fs.v1v6v307.me.raynes.fs/eval2296/fn--2297],
  :as-url
  #function[refactor-nrepl.inlined-deps.fs.v1v6v307.me.raynes.fs/eval2296/fn--2299]}}
```

I suppose having REPL middleware modifying core behaviour could be considered dangerous, though it happens indirectly by`clj-commns/fs`. I shall reach out to them to see what they think.

I originally thought `clerk/file-event`'s `path` was a string and thus the original implementation with passing it to `io/file` and the unit tests written with strings. I suppose in retrospect I should have checked the `path`'s type. and not relied on the REPL#s behaviour...

Thanks!

vemv:
Thanks! Will check if refactor-nrepl actually does use this extension in any way. If not, I might fork the dep and remove the extension.

mk:
Thanks for looking into this again and digging so deep to figure out what the problem is. I also do have refactor-clj on the classpath and hence everything was working for me when I tested it.

Will look into setting up a test suite (with a clear separation of the test and dev class paths) soon.

PR #31, opened 2021-12-14
Option to control opening built static app

filipesilva:
I was trying to build and deploy clerk notebooks as part of https://github.com/athensresearch/athens, but the available build environment there fails because it doesn't have a way to open the built file and `build-static-app!` always tries to open it.

`serve!`, on the other hand, has an option to control this. I thought `build-static-app!` could be controlled in the same way.


mk:
@filipesilva thanks!

PR #36, opened 2021-12-16
Upgrade to Tailwindcss 3 and use via play cdn

zampino:
Upgrade tailwind to version 3 and switch to using the [tailwind play cdn](https://tailwindcss.com/docs/installation/play-cdn).

Also copy the viewer stylesheet in the resource path so it's available with the default deps (the viewers module is only a dev-time dependency).

- [x] use tailwind from CDN
- [x] copy `viewer.css` from viewers in classpath when `bb dev` starts locally


deps.edn
```diff
@@ -18,7 +18,12 @@
         hiccup/hiccup {:mvn/version "2.0.0-alpha2"}
         lambdaisland/uri {:mvn/version "1.11.86"}
 
-        rewrite-clj/rewrite-clj {:mvn/version "1.0.699-alpha"}}
+        rewrite-clj/rewrite-clj {:mvn/version "1.0.699-alpha"}
+
+        ;; needed for CSS jit pass
+        babashka/process {:mvn/version "0.0.2"}
+        io.github.nextjournal/cas {:git/url "git@github.com:nextjournal/cas"
```
zampino, line 25: Damn, this is a private repo :-(

mk: Don’t understand why we would need that. Think we should just have the css locally, and uploading it should be an independent concern, same as we don’t upload the html files.

src/nextjournal/clerk.clj
```diff
@@ -339,15 +346,41 @@
...
-        (browse/browse-url index-html)))))
+        (browse/browse-url index-html)))
+    ;; purge css
+    (when (and purge-css? (not live-js?))
+      ;; ensure we have a local js build
+      (when-not (fs/exists? "build/viewer.js")
+        (when-not (fs/exists? (fs/parent "build/viewer.js"))
+          (fs/create-dirs (fs/parent "build/viewer.js")))
+        (spit "build/viewer.js" ;; slurp latest build from CAS as configured in n.clerk.view
+              (slurp (-> (io/file "src/nextjournal/clerk/view.clj")
```
mk, line 361: There’s a much easier way to get to this, isn’t there? 😜

joe-loco:
There seems to be styles missing when running `bb dev` and looking at the devcards. Am I holding it wrong?

CleanShot 2021-12-17 at 11 33 44@2x

joe-loco:
I tested static builds with and without purge and the results look good. @mk I see what you mean with the waiting times now when `purge-css?` is enabled. I talked with @kommen about this but we both don’t see a way right now to make this more performant.

mk:
Devcards look good here. `yarn install` to the rescue?
CleanShot 2021-12-17 at 12 03 13@2x

zampino:
Just a side note on 

> Consolidate into one tailwind config

in principle, now that we use tailwind from CDN, we'll only need a config in the invocation

```
npx tailwindcss --input stylesheets/app.css --config tailwind-jit.config.js --output public/build/clerk.css -m
```

within `build-static-app!`. We can move such a config into resources and interpolate its absolute path in the command. Same could happen for the `--input` I guess, after @joe-loco refactoring of `viewer.css`.

zampino:
This `.not-prose` stuff is biting us here a bit (but not so crucial), if we use `clerk/html` to render hiccup which actually contains prose (as build from markdown ->hiccup) the [container is assigned a `not-prose` class](https://github.com/nextjournal/clerk/blob/95fcf65fe2567cc3d5c13a993ec6426193ff3c8f/src/nextjournal/clerk/sci_viewer.cljs#L83) and this now happens: 

CleanShot 2022-01-13 at 14 27 32@2x

while it was like this before:

CleanShot 2022-01-13 at 14 27 47@2x

PR #43, opened 2021-12-30
Add `halt!` to stop Clerk

SneakyPeet:
It would be great to be able to stop clerk via the repl using a single command. This PR achieves that by

1. Extracting the webserver stop code out of `nextjournal.clerk.webserver/start!` into the `stop!` function
2. Extracting the beholder stop code out of `nextjournal.clerk/serve!` into the `stop-watcher!` function
3. Adding a new function `nextjournal.clerk/halt!` that calls both functions from 1 and 2

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -239,6 +239,12 @@
 
 #_(file->viewer "notebooks/rule_30.clj")
 
+(defn- stop-watcher! []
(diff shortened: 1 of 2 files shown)


mk:
Hey @SneakyPeet, sorry it took a while to get around to this but it's now merged after some minor changes. Thanks for your PR!

PR #46, opened 2022-01-03
test to cover circular dependency logic

philomates:


test/nextjournal/clerk/hashing_test.clj
```diff
@@ -122,3 +122,14 @@
...
+(deftest circular-dependency
+  (is (match? {:graph {:dependencies {'(ns circular) any?
+                                      'circular/b #{clojure.core/str 'circular/a+circular/b}
+                                      'circular/a #{clojure.core/str 'circular/a+circular/b}}}
+               :->analysis-info {'circular/a any?
+                                 'circular/b any?
+                                 'circular/a+circular/b {:form '(do ((str "boom " b)) ((str a " boom")))}}}
+              (h/analyze-file {:markdown? true}
+                              {:graph (dep/graph)}
+                              "resources/tests/circular.clj"))))
```
mk, line 135: The one-arity version of `analyze-file` should also work here as we're not interested in the (markdown) prose parts.

Also think it might be nice if we could have a way to inline the test notebook code so we can have it more local.

mk (commented):
Thanks for this! The test file locality would be nice to have for me but also fine to merge it without it.

test/nextjournal/clerk/hashing_test.clj
zampino: I like the `with-redefs` trick for testing :-)

> delay that work until the need for handling things like ... 

I'm working on importing clerk files into nextjournal: at present when uploading files like jupyter or markdown we are slurping into a string first and we then apply suitable conversions. It's fine to work with uploaded files ftm but could be useful to be able to clerk-parse a string to stay closer to existing imports.

PR #47, opened 2022-01-06
Make evaluation/cache return variables instead of what they point to (for def/defn)

philomates:
We'd like to allow for different ways to run and render tests from within notebooks.
`clojure.test`'s `deftest` defines a variable with a `:test` metadata field present. It would be great to have this at render time, but clerk currently resolves variables returned by `def` and `defn` into what they point to. The changes here defer this resolution to the viewer so that we can add a viewer that checks for test meta-data and either hides or executes and shows results. We still don't want to resolve normal vars (that are the result of non-def/defn expression evaluation), so I introduced a `wrapped-var` to distinguish between the two.

The PR also:
 - adds test coverage for `read+eval-cached`
 - breaks up `read+eval-cached` into smaller functions. Let me if I went too overboard here; many well-named and small functions helps me with code readability but that might be the case for everyone.

open issues to be resolved:
 - rendering definition that returns a different var is broken
 - rendering `deftest` results in a string result when it should be render a function

20220106_16h18m54s_grim

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -46,7 +46,6 @@
   ;; TODO: validate hash and retry or re-compute in case of a mismatch
   (nippy/thaw-from-file (->cache-file hash)))
 
-
 #_(thaw-from-cas (hash+store-in-cas! (range 42)))
 #_(thaw-from-cas "8Vv6q6La171HEs28ZuTdsn9Ukg6YcZwF5WRFZA1tGk2BP5utzRXNKYq9Jf9HsjFa6Y4L1qAAHzMjpZ28TCj1RTyAdx")
 
@@ -72,9 +71,11 @@
 (defn- lookup-cached-result [results-last-run introduced-var hash cas-hash visibility]
   (try
     (let [value (or (get results-last-run hash)
-                    (thaw-from-cas cas-hash))]
-      (when (and introduced-var (not (::var-from-def value)))
-        (intern *ns* (-> introduced-var symbol name symbol) value))
+                    (let [cached-value (thaw-from-cas cas-hash)]
+                      (prn :intro introduced-var :val cached-value)
+                      (when introduced-var
+                        (intern *ns* (-> introduced-var symbol name symbol) cached-value))
+                      cached-value))]
       (wrapped-with-metadata (if introduced-var (var-from-def introduced-var) value) visibility hash))
     (catch Exception _e
       ;; TODO better report this error, anything that can't be read shouldn't be cached in the first place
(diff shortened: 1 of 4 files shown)


src/nextjournal/clerk.clj
```diff
@@ -58,53 +58,75 @@
...
+(defn- eval+cache! [form hash digest-file introduced-var no-cache? visibility]
+  (let [{:keys [result]} (time-ms (binding [config/*in-clerk* true] (eval form)))
+        var-value (cond-> result (var? result) deref)
+        no-cache? (or no-cache?
+                      (config/cache-disabled?)
+                      (view/exceeds-bounded-count-limit? var-value))]
+    (when (and (not no-cache?) (cachable-value? var-value))
+      (cache! digest-file var-value))
+    (let [blob-id (cond no-cache? (view/->hash-str var-value)
+                        (fn? var-value) nil
```
philomates, line 98: why do functions get no hash, but other non-cachable things do get hashes?

mk:
The printing issues should be resolved with dd19391:
CleanShot 2022-01-07 at 11 24 18@2x

Dropping the `:render-fn` key solved it and the `:transform-fn` doesn't need to be quoted.

src/nextjournal/clerk.clj
mk: because Clojure functions cannot be (de-)serialized.

src/nextjournal/clerk/viewer.cljc
```diff
@@ -351,7 +355,8 @@
    (let [{:as opts :keys [viewers path offset]} (merge {:offset 0} opts)
          wrapped-value (try (wrapped-with-viewer xs viewers) ;; TODO: respect `viewers` on `xs`
                             (catch #?(:clj Exception :cljs js/Error) _ex
-                              nil))
+                              (do (println _ex)
```
philomates, line 358: adding this print helped me catch an issue that required https://github.com/nextjournal/clerk/pull/47/commits/adc43c4e4f094eaa0e4abf6c0ed824a18cee0c2a and then the https://github.com/nextjournal/clerk/pull/47/commits/6dea81e0225d08fa63fc6eccd24c6a9321754b3a follow-up

src/nextjournal/clerk.clj
```diff
@@ -58,53 +58,83 @@
...
+                           :else
+                           (throw (ex-info "Unable to resolve into a variable" {:data var})))]
+    {::var-from-def resolved-var}))
+
+(defn- lookup-cached-result [results-last-run introduced-var hash cas-hash visibility]
+  (try
+    (let [value (or (get results-last-run hash)
+                    (thaw-from-cas cas-hash))]
+      (when (and introduced-var (not (::var-from-def value)))
+        (intern *ns* (-> introduced-var symbol name symbol) value))
```
philomates, line 77: the `results-last-run` for `(def x 1)` will point to `{::var-from-def #'x}` and then without the `(not (::var-from-def value))` we would `(intern *ns* #'x {::var-from-def #'x})` which causes the viewer code to stackoverflow when it it tries to display this var.

I'm unsure what the best approach to caching vars that we want to leave unresolved until viewer time (to allow for test running for example). So I just added a skip here. I think that the cache-to-filesystem code still caches the var->value even if it comes from a def, we just never use it.

PR #49, opened 2022-01-10
proof-of-concept: example notebook that runs clojure.test tests

philomates:
proof of concept PR that uses the changes in https://github.com/nextjournal/clerk/pull/47 to allow for adding a viewer that runs vars marked with `:test` metadata

20220110_12h37m36s_grim

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -42,6 +42,7 @@
            :demo {:extra-deps {com.github.seancorfield/next.jdbc {:mvn/version "1.2.659"}
                                org.xerial/sqlite-jdbc {:mvn/version "3.34.0"}
                                org.clojure/data.csv {:mvn/version "1.0.0"}
+                               hickory/hickory {:mvn/version "0.7.1"}
                                sicmutils/sicmutils {:mvn/version "0.20.0"}}}
 
            :build {:deps {io.github.clojure/tools.build {:git/tag "v0.6.1" :git/sha "515b334"}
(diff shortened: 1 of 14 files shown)


deps.edn
```diff
@@ -1,5 +1,5 @@
 {:paths ["src" "resources"]
- :deps {org.clojure/clojure {:mvn/version "1.10.3"}
+ :deps {org.clojure/clojure {:mvn/version "1.11.0-alpha3"}
```
philomates, line 2: to get `clojure.test/run-test-var`

notebooks/test_driven.clj
```diff
@@ -0,0 +1,42 @@
...
+    (str s)))
+
+(defn test-runner-viewer [v]
+  (with-test-out->str #(t/run-test-var v)))
+
+
+(clerk/set-viewers!
+  [{:pred         test-var?
+    :transform-fn #(-> % :nextjournal.clerk/var-from-def test-runner-viewer)
+    :render-fn    #(v/html [:span.syntax-string.inspected-value %])}])
```
philomates, line 27: instead of the default string viewer, to prevent shortening of output. 

todo: make linebreaks show correctly

notebooks/test_driven.clj
```diff
@@ -0,0 +1,42 @@
...
+
+(deftest foo
+  (is (= 1 2))
+  (is (= 1 1)))
+
+;; if we want to display `testing` results we'll need to wrap them somehow like
+;; this:
+(with-test-out->str
+  #(testing
+     (is (= [1] [2]))))
```
philomates, line 37: we could potentially explore the ability to inline `testing` expressions and have clerk see them and run them just like it runs `deftest`s here. Might be a bit ergonomic, but is also a bit messy because top-level `testing` will always print test failures on ns reloads

PR #50, opened 2022-01-11
Some viewer tests + fix unnecessary render-fn wrapping after ns reload

philomates:


src/nextjournal/clerk/viewer.cljc
```diff
@@ -17,9 +17,13 @@
...
 
-
 (defn fn+form? [x]
-  (instance? Fn+Form x))
+  (= (str Fn+Form)
+     (str (type x))))
+
+(defn form? [x]
+  (= (str Form)
+     (str (type x))))
```
philomates, line 26: not sure if this is the best way to check for instance of a record, but using `instance?` doesn't work well with namespace reloads. Reloading the namespace would change the results to wrap the render-fn again due to it not being an instance of the re-created `Form` defrecord: https://github.com/nextjournal/clerk/pull/50/commits/dadca3a475f6520b58cb34608a8b37ad5d5967de#diff-8479f3457f1323fbfe27364cdf11308963c82e8e2803ae0781c438c4ae049b4cL84-L87

PR #55, opened 2022-01-17
Allow to bring your own css and js

mk:
This lets consumers opt into providing their own js and js by setting values in `nextjournal.clerk.config/ !resource->url`. Also drop `live-js?` the config but introduce a `clerk.resource_manifest` JVM system property and use it to get live js in dev instead.

### DIFF
--- a/build.clj
+++ b/build.clj
@@ -41,14 +41,13 @@
 
 (defn asserting [val msg] (assert (seq val) msg) val)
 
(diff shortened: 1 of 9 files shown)


genmeblog:
This PR makes that CSS is replaced completely. On the other hand it would be great to just enhance Clerk's CSS. For example: I want to add two or three custom classes only.

PR #57, opened 2022-01-18
Refactor to avoid double parsing and make testing easier

mk:
This refactors the hashing and eval to:
- only parse & analyze each form once fixing an `Could not resolve var` error that could previously occur in the second analysis pass
- also provide string based interfaces and use for testing instead of going through files
- move Clerk's cache into the `.clerk` folder
- drop metadata and use a map to make conveyance of previous results more explicit

### DIFF
--- a/src/nextjournal/clerk/hashing.clj
+++ b/src/nextjournal/clerk/hashing.clj
@@ -375,12 +375,10 @@
 #_(dep/transitive-dependencies (:graph (build-graph "src/nextjournal/clerk/hashing.clj"))  #'nextjournal.clerk.hashing/long-thing)
 
 (defn- hash-form [->hash {:keys [hash form deps]}]
-  (prn :hash-form ->hash :hash hash :form form :deps deps)
   (let [hashed-deps (into #{} (map ->hash) deps)]
     (sha1-base58 (pr-str (conj hashed-deps (if form form hash))))))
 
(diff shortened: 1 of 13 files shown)


src/nextjournal/clerk/hashing.clj
```diff
@@ -237,9 +238,18 @@
               deps)
       state-with-var-hash)))
 
+(defn analyze-doc
+  ([doc]
+   (analyze-doc {:graph (dep/graph)} doc))
+  ([state doc]
+   (let [state-with-document (assoc state :doc doc)
+         code-cells (into [] (filter (comp #{:code} :type)) (:doc doc))]
```
philomates, line 246: Probably worth lifting `(into [] (filter (comp #{:code} :type)) (:doc doc))` into a helper and calling both here and from `analyze-file`

philomates: should `analyze-doc` ever care about the `:markdown?` option? If so it might be worth writing `analyze-file` read the file and pass it to `analyze-doc` for doing work on

philomates (approved):

src/nextjournal/clerk/hashing.clj
zampino, line 246: Maybe inline bindings above?

zampino (approved):
Looks good to me!

PR #62, opened 2022-01-27
Overridable Markdown node Viewers

zampino:
This set of changes (partially <sup>1</sup>) fits markdown rendering into the viewers API framework, in doing so we achieve: 

* each markdown node (`:heading`, `:paragraph`, `:monospace`, etc) gets assigned an individual viewer and as such its appearance is completely customisable by the user via `set-viewers!` function. Check the [markdown viewers notebook](https://snapshots.nextjournal.com/clerk/build/b1b010e26856c0b8ba20085f797802fd241d9509/index.html#/notebooks/viewers/markdown.clj).
* markdown viewers might depend on other viewers. For instance the formula viewer now is implemented  on top of the `:latex` viewer: changing the latter will affect all formulas inserted in markdown fragments. Check the [mathjax example](https://snapshots.nextjournal.com/clerk/build/b1b010e26856c0b8ba20085f797802fd241d9509/index.html#/notebooks/viewers/mathjax.clj).

It is debatable if this is sufficient or if we want to move viewer selection on the clojure side as per latest trend.

- [ ] fix safari issues with `Error: MathJax retry`

#### Notes
<sup>1</sup>: Markdown blocks are not described but passed to the client wrapped in `with-viewer` format. The client is still responsible for looking up the right viewer for each markdown node.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -25,7 +25,7 @@
            :sci {:extra-deps  {applied-science/js-interop  {:mvn/version "0.3.0"}
                                borkdude/sci {:mvn/version "0.2.6"}
                                reagent/reagent {:mvn/version "1.1.0"}
-                               io.github.nextjournal/viewers {:git/sha "3e1569a67dbec87bcbd7d1343aa754222aeede77"}
+                               io.github.nextjournal/viewers {:git/sha "fd4a1b9ec220e17d2893988c57be2a239d165ef1"}
                                metosin/reitit-frontend     {:mvn/version "0.5.15"}}}
 
            :dev {:extra-deps {arrowic/arrowic {:mvn/version "0.1.1"}
--- a/notebooks/mathjax.clj
+++ b/notebooks/mathjax.clj
@@ -1,22 +1,21 @@
-(ns mathjax
+;; # MathJax
+(ns ^:nextjournal.clerk/no-cache mathjax
   (:require [nextjournal.clerk :as clerk]
             [nextjournal.clerk.viewer :as clerk-viewer]))
 
-#_(clerk/set-viewers! [{:name :latex :render-fn (quote v/mathjax-viewer) :fetch-fn (fn [_ x] x)}])
-#_(reset! clerk-viewer/!viewers clerk-viewer/all-viewers)
+(clerk/set-viewers! [{:name :latex
+                      :render-fn (quote v/mathjax-viewer)
+                      :fetch-fn (fn [_ x] x)}])
 
-(swap! clerk-viewer/!viewers update :root (partial mapv #(cond-> %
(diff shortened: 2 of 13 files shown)


zampino:
Trying to do viewer selection on the clojure side but it's not looking good. Starting with just a few nodes with a recursive describe step done via `:transform-fn` (which doesn't get passed describe `opts`) and unwrapping via `value` fn:

```clojure

(def describe-children (map (comp value #(describe % {:viewers default-markdown-viewers}))))

(def default-markdown-viewers
  [{:name :nextjournal.markdown/formula
    :fetch-fn (fn [_ {:keys [text]}] text)
    :render-fn (quote v/katex-viewer)}

   {:name :nextjournal.markdown/text
    :fetch-fn (fn [_ {:keys [text]}] text)}

   {:name :nextjournal.markdown/paragraph
    :fetch-fn fetch-all
    :transform-fn (fn [{:nextjournal/keys [value]}]
                    (into [:p] describe-children (:content value)))}

   {:name :nextjournal.markdown/heading
    :fetch-fn fetch-all
    :transform-fn (fn [{:nextjournal/keys [value]}]
                    (let [{:keys [heading-level content]} value]
                      (into [(keyword (str \h heading-level))]
                            describe-children
                            content)))}])

(def default-viewers
  [{:name :nextjournal.markdown/doc
    :render-fn 'v/html
    :transform-fn (fn [wrapped-value]                       ;; FIXME: transform-fn doesn't take options
                    (-> wrapped-value
                        (update :nextjournal/value (fn [{:keys [content]}]
                                                     (into [:<>]
                                                           describe-children
                                                           content)))))
    :fetch-fn fetch-all}
...
```
and this looks kind of ok

```clojure
(-> (nextjournal.markdown/parse "# Hello
some paragraph $\\phi$.
  ")
      nextjournal.clerk.view/md->viewer
      describe)
;; => 
{:path [],
 :nextjournal/value [:<> [:h1 "Hello"] [:p "some paragraph " "\\phi" "."]],
 :nextjournal/viewer {:name :nextjournal.markdown/doc, :render-fn #viewer-fnv/html}}
```
but I cannot easily get a grasp on the formula node `:render-fn` to have it use the latex viewer. We'd also have to check that this approach is still open for user-facing overrides.

zampino:
> or if we want to move viewer selection on the clojure side

This is actually relying on "client-side viewer registration" which is being dropped with https://github.com/nextjournal/clerk/compare/drop-eval-viewer. The current approach won't make sense any longer, so we'll need to go for an actual server-side description.

zampino:
Superseded by https://github.com/nextjournal/clerk/pull/122.

PR #72, opened 2022-02-03
catch and rethrow eval error with additional context

philomates:
It can be hard to track down the source of notebook evaluation errors. This PR adds the original file name and form for which the evaluation exception is relevant.

for a notebook with an expression like `(throw (ex-info "boom" {:hi :there}))` we have:

## before change
20220203_14h37m04s_grim

## after change

20220203_14h36m22s_grim


daveliepmann:
you're doing god's work 🙇

mk:
Sorry that I never looked at this and instead did a similar thing with e5c00bb0473bfdf1562024879a5ff69ed3309477.

PR #81, opened 2022-02-15
added support to freeze fns

### LINKED ISSUE #80: allow freeze/ unfreze of fns
As discussed in slack:
https://app.slack.com/client/T03RZGPFR/C01GH5EN7JA/thread/C01GH5EN7JA-1644827215.944249

This library allows to do so:

https://github.com/redplanetlabs/nippy-serializable-fns

behrica:
fixes #80

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -12,6 +12,8 @@
         io.github.nextjournal/markdown {:mvn/version "0.2.44"}
 
         com.taoensso/nippy {:mvn/version "3.1.1"}
+        com.rpl/nippy-serializable-fns {:mvn/version "0.4.1"}
+
         mvxcvi/multihash {:mvn/version "2.0.3"}
         javax.xml.bind/jaxb-api {:mvn/version "2.3.1"} ;; needed to make multihash work on JDK11+
 
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -12,7 +12,9 @@
             [nextjournal.clerk.view :as view]
             [nextjournal.clerk.viewer :as v]
             [nextjournal.clerk.webserver :as webserver]
-            [taoensso.nippy :as nippy])
+            [taoensso.nippy :as nippy]
+            [com.rpl.nippy-serializable-fn] ;; enable freeze of fns
+            )
   (:import (java.awt.image BufferedImage)
            (javax.imageio ImageIO)))
 


jackrusher:
@behrica Did you test this across restarts of your JVM?

behrica:
My current example code has an other freeze / unfreeze issue, which forces me to delete the clerk cache on JVM start.
So I did not test it.

behrica:
I tested with this code:

```
(ns test

  (:require [nextjournal.clerk :as clerk]
            [tablecloth.api :as tc]
            [nextjournal.clerk :as clerk]
            [nextjournal.clerk.hashing :as h]
            [weavejester.dependency :as dep]))

(comment
  (require '[com.rpl.nippy-serializable-fn])
  (clerk/clear-cache!)
  (clerk/serve! {:browse? true})
  (clerk/serve! {:watch-paths ["notebooks" "src"]}))

(defn my-fn [] nil)

(def b {:fn my-fn
        :y (do (println "slow") (Thread/sleep 10000) :a)})
```

and it worked. 

After JVM restart, and (clerk/show! ...)  the "slow" did not get printed, and the evaluation was fast.
-> `b` came from cache

jackrusher:
The test case you've made here uses a bound function of no arguments that doesn't make reference to any outside variables in its closure. For this to be a general solution, it would also need to work on (for example) anonymous functions that do close over bindings. The announcement blog post for the library you've added here includes, in the comments at the bottom, this quote:

> All JVM instances that could end up deserializing a fn instance are required to be launched from the same precompiled jar
https://tech.redplanetlabs.com/2020/01/06/serializing-and-deserializing-clojure-fns-with-nippy/

Because of these limitations, I think we'll end up with a solution that uses in-memory caching for functions, but which does not try to write them to disk.

mk:
I took a stab at an alternative approach in #82. This does not change the types that nippy can serialize but instead fixes a but that Clerk did not use the already existing in-memory cache for results that were not freezable using nippy. this has the upside of working for any result (as you've noted, there's a lot more types besides Clojure functions that nippy will not serialize by default). Please let me know if this works for you.

mk:
Going to close this in favor of #82 for now. I'd like to better understand the limitations of this before considering it. Clerk's nippy cache is really meant to store results across for longer periods and across JVM restarts, conditions for which we know the guarantee that the code backing the functions must not change can certainly be violated.

That said, it's possible to bring in this dependency and play with this in users space without requiring modifications to Clerk and I encourage you or other folks to do this and report back.

PR #87, opened 2022-02-19
log message for freeze error

behrica:
It just prints one line


mk:
I do believe surfacing these errors makes sense. Rather than using the REPL output streams we want to use the browser window to surface information like this, we're tracking this in https://github.com/nextjournal/clerk/blob/7ddbc3a1b6d427b306b73f7e7c93e4fe03282d91/notebooks/onwards.md?plain=1#L95

PR #88, opened 2022-02-19
Update how_clerk_works.clj

behrica:
clarify caching


mk:
This should become documented inline once we're tackling https://github.com/nextjournal/clerk/blob/7ddbc3a1b6d427b306b73f7e7c93e4fe03282d91/notebooks/onwards.md?plain=1#L95

PR #90, opened 2022-02-20
delete in-memory cache on clean

behrica:


mk:
I started working on this in afa618f4e3f5404639b25d8882274b4f969f496b but discovered more issues (the frontend not seeing updates without a reload as the hash didn't change) along the way that will require a larger refactoring.

behrica:
I saw it. I close the PR.

Issue #66, opened 2022-01-30
dependency resolution in case of double defined defs is non-deterministic


behrica:
Code such as 

```clojure
(def a 1)
...
...
(def a 2)
(+ a 1)
```
in an notebook results sometimes in 2 , sometimes in 3

behrica:
Code like thgis coukld detect such as situation:

```
(def parsed
  (clerk/parse-file "notebooks/duplicate.clj"))


(def analyzed
  (h/build-graph parsed))

(def duplicate-vars
  (->> analyzed :doc :blocks (map :var) frequencies (filter (fn [[key val]] (>  val 1))) keys vec))
```

I propose to run such a detection during analysis, and fail on it for now, saying:

```
"Clerk's dependency analysis and caching cannot handle re-refined var's"
The following vars are redefined ....
```

Having this in a namespace can result in un-noticeable and hard to debug issues.

PR #92, opened 2022-02-21
check and fail on duplicated defs

behrica:
addresses #66


mk:
Sorry for leaving this dangling for so long. We'll want to take a different approach and support redefining vars in the future.

PR #97, opened 2022-02-22
Nbb UI tests: first cut

borkdude:


.github/workflows/main.yml
```diff
@@ -102,3 +107,8 @@ jobs:
           state: 'success'
           sha: ${{github.event.pull_request.head.sha || github.sha}}
           target_url: https://snapshots.nextjournal.com/clerk/build/${{ github.sha }}
+
+      - name: Run Playwright tests against static assets
+        run: |
+          bb test:static-app ${{ github.sha }}
```
mk, line 113: Can we make this run as a separate task, producing a separate GitHub status badge? I'd like to be able to differentiate between a build failure of the static app vs a failure of the browser tests.

nbb/package-lock.json
```diff
@@ -0,0 +1,347 @@
+{
```
mk, line 1: This is a lock file from npm, but we're calling `yarn install`, leading to this warning:
> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

nbb/playwright_tests.cljs
```diff
@@ -0,0 +1,143 @@
...
+         (def ~name res))
+       (.catch js/console.log)))
+
+(comment
+  (launch-browser)
+  (defp p (.newPage @browser))
+  (.on p "console" (fn [msg]
+                     (when (= "error" (.type msg))
+                       (swap! console-errors conj msg))))
+  (goto p "https://dude.devx")
```
mk, line 130: Dude?

bb/utils.clj
```diff
@@ -0,0 +1,6 @@
+(ns utils
+  (:require [babashka.process :refer [sh]]
+            [clojure.string :as str]))
+
+(defn latest-sha []
+  (str/trim (:out (sh "git rev-parse HEAD"))))
```
mk, line 6: My preference would be to inline this in `bb.edn` to not have another extra dir & file.

mk (commented):
The general approach here is to browse to all notebooks in the static build and assert they don't log any console errors, correct?

Did you verify that viewer errors are logged and lead to a failure or should we still do that?

borkdude:
> The general approach here is to browse to all notebooks in the static build and assert they don't log any console errors, correct?

Correct.

> Did you verify that viewer errors are logged and lead to a failure or should we still do that?

I verified this in the REPL, but we can verify this by making a test commit which should trigger this. Do you know of anything that should trigger a console error in a clerk notebook?

.github/workflows/main.yml
borkdude: Done

nbb/package-lock.json
borkdude: There is no `npx` in yarn. Why is nextjournal using yarn instead of the default node tooling?

borkdude: It seems `yarn nbb` also works. TIL.

nbb/playwright_tests.cljs
borkdude: Test code to trigger a console error.

bb/utils.clj
borkdude: This can lead two ways: big chunks of code in bb.edn, making it, as you pointed out once, less REPL friendly. Or we start right now with a separate file for developing larger chunks of code and move the code that is already in `:init` to this file.

borkdude:
@mk Addressed everything, except one comment about the extra dir. Imo it's better to start coding in a separate file and use bb.edn only for hooking up code to task names, but given that the code is relatively thin, I'd be ok with moving it to the bb.edn for now. Just let me know.

borkdude:
Hmm, somehow removing package lock json causes an error, I'll look into it.

borkdude:
Running `bb test:static-app bee6aa1bde9dfe88828171b1217e1c596d402ad9` locally works fine though.

borkdude:
Turned out to be a timeout problem.

borkdude:
Removing bb dir for now.

borkdude:
Make a commit to trigger a UI error (hopefully). Don't merge yet.

mk:
Need to `bb release:js` and push that for the cljs change to have an effect.

borkdude:
OK, should I do this in the build step?

mk:
Was about to merge this until I saw that you added an explicit `js/console.error` in 0465c742cc1bbba961291f90f7121ef288099699. So I'd like to verify that the tests also catch other failure modes (sci read or eval errors, as well as runtime exceptions). I plan to do this first thing tomorrow and then ship it.

borkdude:
I think you made a mistake with https://github.com/nextjournal/clerk/pull/97/commits/f71bcb47b315eac7fbbb0be9e1bf23dbf75341f0 since the tests contain a static build test which needs the JS asset to already be present.

mk:
@borkdude you mean this one? https://github.com/nextjournal/clerk/blob/e031d5135e49f4816a9b472f528d5023c02d9582/test/nextjournal/clerk_test.clj#L17-L27

If yes that should not be a problem since it's not asserting anything about this. In general, I'd like to keep the fast feedback loop the clojure tests currently bring and would prefer it if we managed to keep them without a dependency on the viewer js build.

mk:
Oh, I see what you mean now. The slurp will fail if the js hasn't been compiled yet. I guess we can pass an override setting for the tests.

PR #98, opened 2022-02-24
Support non-root git location in static builds link

grahamcarlyle:
Added a `:deps/root` optional parameter to static build function to link to a deps project at a non-root git location. Chose `:deps/root` to mirror the deps.edn dependency coordinate attribute for a non-root file location.


mk:
Sorry for leaving this in a dangling state for years. Will probably want to support this using a more open extension mechanism using a function.

PR #107, opened 2022-03-07
Write hash in pre-commit hook for predicable viewer.js location, build JS in CI

### LINKED ISSUE #102: Include file in repo that contains CAS-hash to JS asset
This solves the problem of not having to change the source in CI when building the JS.

- This hash can be calculated fast using dejavu + bb and will be populated on every commit using a git pre-commit hook
- The JS asset itself will be built on CI and uploaded using this hash
- For local development, the hash will be ignored and the locally built JS will be used.

This change is necessary to reliably test changes in the UI, see #97 

/cc @mk 

TODO:

- [ ] upload static asset via dejavu using 512 hash
- [ ] update lookup link to point to 512 hash

borkdude:
Fixes #102 

@mk Ready for review.


bb.edn
```diff
@@ -18,13 +23,22 @@
...
-                               (clojure "-T:build upload-to-cas+rewrite-sha :resource viewer.js"))
-                     :depends [yarn-install]}
+         build:js {:doc "Builds JS"
+                   :depends [yarn-install]
+                   :task (clojure "-M:sci:demo:dev release browser")}
+
+         write-hash {:doc "Write hash to resources. Intended to be used in a pre-commit hook."
+                     :task hashing/write-hash}
+
+         refresh-cdn {:doc "Refreshes assets stored on CDN (google storage)"
```
mk, line 33: It's not clear to me from the name or doc what this tasks does. Does it build the viewer.js and upload it to google cloud storage?

mk (commented):
Thank you, this looks good and getting this should nicely improve the workflow. Only wondering if we can do a pass on the names here to make it clearer what this is for.

Some ideas:
* `write-hash` → `write-viewer-resources-hash`
* `hashing.clj` → `viewer-resources-hashing`
* `refresh-cdn` → `build+upload-viewer-resources`
* `front-end-hash.txt` → `viewer-js-hash.txt`, or `viewer-resource-manifest` if we turn it into the same map that `clerk.resource_manifest` is.

Note that in `clerk.resource_manifest` we already use a custom css if [that key](https://github.com/nextjournal/clerk/blob/1d89df7c428616c2c4e12a372692daa3fdd1ec1d/src/nextjournal/clerk/view.clj#L184) is set.

Alternatives to `resources` could be `assets` (which I know you don't like) or to stay as narrow as possible `viewer-js`. Though I do want to add building a css file here very soon.

bb.edn
borkdude: Yes

borkdude:
@mk Excellent suggestions. I also renamed the `bb` dir to `src-bb` which is maybe a better name. I processed all suggestions. We can do more renaming in the future as necessary. I'll fix the build as renaming things almost always is paired with build failures :P.

borkdude:
Oh btw, I think I prefer `refresh-viewer-assets` (or resources) over `build+upload-viewer-resources` since that step doesn't always build, it only builds if the assets need to be refreshed on the CDN, hence `refresh-cdn`. But this is bikeshedding, I'll just go with your name.

borkdude:
@mk Implemented the changes. One thing I'm wondering about. What if we get hash collisions, e.g.

gs://nextjournal-cas-eu/data/lookup/2MnmwsYfak9x5j6cXDBB56uRiheT

already exists, but was produced from a different set of files a year or so ago. Should we do additional cache invalidation?

borkdude:
As discussed: collisions aren't considered a problem. Ready for squash/merge!

bb/viewer_resources_hashing.clj
```diff
@@ -0,0 +1,71 @@
...
+            [nextjournal.dejavu :as djv]))
+
+(def output-dirs ["resources/public/ui"
+                  "resources/public/build"])
+
+;; Example link in bucket:
+;; "https://storage.googleapis.com/nextjournal-cas-eu/data/8VwKauX6JACEP3K6ahNmP5p1w7rWdhKzeGXCDrHMnJiVrUxHVxcm3Xj84K2r3fcAKWxMQKzqoFe92osgFEHCuKCtZC"
+
+(def gs-bucket "gs://nextjournal-cas-eu/data")
+(def base-url "https://storage.googleapis.com/nextjournal-cas-eu/data")
```
mk, line 15: I think we should drop the `/data` prefix from the base-url and gs-bucket and create the lookup files on the top-level.

resources/viewer-js-hash
```diff
@@ -0,0 +1 @@
+yh4majqZ75XpcUpjEFqRovMNjpk
```
mk, line 1: Maybe a simplification / generalization would be to save the full lookup url here?

src/nextjournal/clerk/config.clj
```diff
@@ -9,8 +11,9 @@
   (when-let [prop (System/getProperty "clerk.disable_cache")]
     (not= "false" prop)))
 
-(def default-resource-manifest
-  {"/js/viewer.js" "https://storage.googleapis.com/nextjournal-cas-eu/data/8Vt3Eko1RGonTrRXaLTP6Ly5XugGnYiRapACBxSoZ2RKwFztsi2NYXvTb5vyCTFya2RDgK9vXwVvKBnQ8dWopcmvA5"})
+(def gs-url-prefix "https://storage.googleapis.com/nextjournal-cas-eu/data")
```
mk, line 14: `/data` should also be dropped here.

src/nextjournal/clerk/config.clj
```diff
@@ -19,7 +22,9 @@
 
 (defonce !resource->url
   (atom (or resource-manifest-from-props
-            default-resource-manifest)))
+            ;; assume that CI will have published a CAS-link under this lookup,
+            ;; prior to hitting this code-path
+            (edn/read-string (slurp lookup-url)))))
```
mk, line 27: The one thing I'm struggling with a bit here, is that it's pretty opaque what's inside the lookup file now (without running curl against the lookup-url, that is). Wondering if there's a way for us to get some of the clarify of `default-resource-manifest` back.

mk (commented):
Some remarks which we might want to address in a follow-up.

bb/viewer_resources_hashing.clj
```diff
@@ -0,0 +1,71 @@
...
+  (str gs-url-prefix "/" hash))
+
+(defn build+upload-viewer-resources []
+  (let [front-end-hash (str/trim (slurp viewer-js-hash-file))
+        manifest (str (fs/create-temp-file))
+        res (djv/gs-copy (str (lookup-url front-end-hash)) manifest false)]
+    (when (= res ::djv/not-found)
+      (tasks/run 'build:js)
+      (let [content-hash (djv/sha512 (slurp "build/viewer.js"))
+            viewer-js-http-link (str (cas-link content-hash) "?cache=false")]
```
mk, line 66: Oh, and why do we append `"?cache=false"`?

borkdude: This is because the file was cached and I wrote an invalid one before that.

borkdude:
Thanks, will resolve in follow up.

resources/viewer-js-hash
borkdude: That would not be a simplication because in some cases we need the gs:// url and in other cases we need the http:// url.

src/nextjournal/clerk/config.clj
borkdude: I'll add a comment.

borkdude:
Processed all comments and pushed to the ui tests branch (to avoid merge conflicts, rebasing, etc, I'd like to limit the work in progress).

PR #109, opened 2022-03-10
TOC Redux 🗂 & Dark Mode Toggle 🌗

joe-loco:
This PR brings a table of contents component that is visible by default on deskop. The TOC is opt-in via the `^:nextjournal.clerk/toc` metadata on the ns form (which can be collapsed via `^{:nextjournal.clerk/toc :collapsed}`). In addition to that, dark mode is now also exposed via its own toggle. TOC and dark mode state is persisted in localStorage.

CleanShot 2022-03-30 at 17 08 06

* [x] Use https://github.com/nextjournal/viewers/pull/53 for TOC
* [x] Pin toggle
* [x] Persist in localStorage
* [x] Make scrolling to headings work
* [x] Test edge cases (no headings, only headings of same level, no `h1`, etc.)
* [x] Animate scrolling to headings (and add scroll padding)
* [x] Improve hit targets on mobile
* [x] Option to opt-in
* [x] Adjust for Clerk’s dark mode
* [x] Make static build work with pinned ToC


mk:
@joe-loco we still need to fix the static build here: https://snapshots.nextjournal.com/clerk/build/59dcbf37ac3aa2603b948e6fb826ff38e3818511/index.html#/notebooks/viewer_api.clj

mk:
Two (hopefully) last things I noticed in the [static build](https://snapshots.nextjournal.com/clerk/build/512d17b321c36f0de101287717ff6a4bbe1cf7c2/index.html#/notebooks/viewer_api.clj): 
* sliding in the ToC via hover and then pinning it doesn't scroll the main contents to the right. 
* I'm not sure if it's good that we pin the ToC when you click the hamburger menu. But maybe it's also fine & clear once we fix the first issue.

mk:
See it in action at https://snapshots.nextjournal.com/clerk/build/1b854ae9d2fbc205141a4983393c891b530e56ea/index.html#/notebooks/viewer_api.clj

PR #111, opened 2022-03-11
Set `:value` in interactivity notebook

sritchie:
The `:initial-value` setting on an `:input` element causes react to throw an error saying it will ignore the setting. I think the right way to handle these is to simply set `:value`;

This has the nice effect of setting the initial slider (or textbox) values to the state in the atom, so they persist across reload. Without this change, the text box is blank on initial load, vs displaying the atom source.


mk:
What I did want to set here was the `defaultValue` as to get an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html). This was mainly to improve the typing performance. Think you're right that we should be setting just `:value` and continue to work on the performance.

PR #124, opened 2022-04-01
Remove column truncation

ezmiller:
## Problem

When there are a lot of columns in a table rendered using the `nextjournal.clerk.sci-viewer/table-viewer`, the column names are truncated:

This makes the exploration of datasets with many columns very difficult. 

## (One) Solution

One fix seems to be to remove the tailwind `truncate` class that is being used on the markup for the table, and to modify the CSS slightly to layout the value so it is lined up against the edge of the column name:

image

This way of handling things mimics how, for example, Pandas dataframes are rendered in a python notebook: 

image


ezmiller:
Interestingly, once the column names do not collapse it looks like the number of columns displayed is limited, but not the column names. As seen here:

image

But this also seems a bit out of scope for this PR perhaps...

src/nextjournal/clerk/sci_viewer.cljs
```diff
@@ -481,9 +481,8 @@
                                       :title (if (or (string? k) (keyword? k)) (name k) (str k))}
                                      [:div.inline-flex
                                       ;; Truncate to available col width without growing the table
-                                      [:div.table.table-fixed.w-full.flex-auto
-                                       {:style {:margin-left -12}}
```
ezmiller, line 485: Hmmm, looking at examples in `notebooks/viewers/table.clj` this seems to align certain columns correctly. Most of the tables look weird without this. Eg.:
image
But then my test looks strange with it: 
image

Issue #121, opened 2022-03-24
Can't define a record named `Node`


jprudent:
```clojure
(defrecord Node [v l r])
(->Node 1 nil nil)
```
will fail with the following error:

```
Unhandled clojure.lang.ArityException
Wrong number of args (3) passed to: clojure.lang.PersistentArrayMap

AFn.java: | 429 | clojure.lang.AFn
AFn.java: | 40 | clojure.lang.AFn
NO_SOURCE_FILE: | 0 | day-0003/eval16646 // <-- my journal
NO_SOURCE_FILE: | -1 | day-0003/eval16646
Compiler.java: | 7181 | clojure.lang.Compiler
Compiler.java: | 7136 | clojure.lang.Compiler
core.clj: | 3202 | clojure.core/eval
core.clj: | 3198 | clojure.core/eval
clerk.clj: | 104 | nextjournal.clerk/eval+cache!/fn--15074
clerk.clj: | 104 | nextjournal.clerk/eval+cache!
clerk.clj: | 103 | nextjournal.clerk/eval+cache!
clerk.clj: | 155 | nextjournal.clerk/read+eval-cached
clerk.clj: | 129 | nextjournal.clerk/read+eval-cached
clerk.clj: | 177 | nextjournal.clerk/eval-analyzed-doc/fn--15108
PersistentVector.java: | 343 | clojure.lang.PersistentVector
core.clj: | 6829 | clojure.core/reduce
core.clj: | 6812 | clojure.core/reduce
```

Totally fine with
```clojure
(defrecord Nod [v l r])
(->Nod 1 nil nil)
```

And works great in a REPL
```
;;Loading daily_coding_problem/day_0003.clj... done
(in-ns 'day-0003)
=> #object[clojure.lang.Namespace 0x7dfb9dd "day-0003"]
(->Node 1 2 3)
=> #day_0003.Node{:val 1, :left 2, :right 3}
```

Further in my session I got this error
```
Multiple definitions found in form, using first one:
:form (defrecord Nod [val left right]) :used-var day-0003/->Nod
```

So I suppose this something weird about records in general.

philomates:
Hi there,
I wasn't able to reproduce the `Unhandled clojure.lang.ArityException, Wrong number of args (3) passed to: clojure.lang.PersistentArrayMap` issue, but https://github.com/nextjournal/clerk/pull/128 is a potential way forward to the `Multiple definitions found in form, using first one:` warning

mk:
[...]

PR #128, opened 2022-04-07
Analyze defrecords as Name, ->Name, map->Name

philomates:
potential approach to fixing the warning mentioned in https://github.com/nextjournal/clerk/issues/121 where
```clojure
(defrecord Node [v l r])
```
results in Clerk outputting
```
Multiple definitions found in form, using first one:
:form (defrecord Node [val left right]) :used-var day-0003/->Node
```
because `(defrecord Node ...)` macroexpands to several `defn` forms including `->Node` and `map->Node`

This PR 
 - adapts the analysis to include a `defrecord` special case where the defrecord name is used instead of the first `defn` form.
 - creates an explicit dependency from the defrecord name to `->RecordName` and `map->RecordName` such that if the record body changes, any usages of `->RecordName` or `map->RecordName` will register as needing to be re-evaluated

Not sure if such a special case is acceptable or if we should re-do the dependency analysis to more holistically deal with expressions with multiple `def`s in them

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,5 +1,5 @@
 {:paths ["src" "resources"]
- :deps {org.clojure/clojure {:mvn/version "1.10.3"}
+ :deps {org.clojure/clojure {:mvn/version "1.11.1"}
         org.clojure/java.classpath {:mvn/version "1.0.0"}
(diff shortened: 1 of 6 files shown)


mk:
Thanks, a more generic approach in #159 went surprisingly well so going with that instead.

PR #129, opened 2022-04-07
`viewer-for` helper

philomates:
Sometimes it is hard to know what viewer will be applied to a value.
You can get this info from `nextjournal.clerk.viewer/describe` but it also runs the `transform-fn` and recursively applies viewer functions into nested values.

I think it might be useful for development to have a construct that looks up the top level viewer for a value without considering nested values or performing transformations. In this PR I implemented it as `viewer-for`

Note that it seems to deviate a little bit as compared to the viewer that `describe` returns:

```clojure
((comp :nextjournal/viewer describe) [1])
;; =>
{:render-fn {:form v/coll-viewer, :f nil},
 :opening-paren "[",
 :closing-paren ("]"),
 :fetch-opts {:n 20}}
```
while

```clojure
(viewer-for [1])
;; =>
{:pred #function[clojure.core/vector?--5479],
 :render-fn v/coll-viewer,
 :opening-paren "[",
 :closing-paren "]",
 :fetch-opts {:n 20}}
 ```

### DIFF
--- a/src/nextjournal/clerk/viewer.cljc
+++ b/src/nextjournal/clerk/viewer.cljc
@@ -372,24 +372,26 @@
   (when (wrapped-value? x)
     (select-keys x [:nextjournal/width])))
 
+(defn- lookup-viewer-for [x viewers]
+  (let [v (value x)]
+    (if-let [selected-viewer (viewer x)]
+      (if (keyword? selected-viewer)
+        (or (find-named-viewer viewers selected-viewer)
+            (throw (ex-info (str "cannot find viewer named " selected-viewer)
+                            {:selected-viewer selected-viewer :x v :viewers viewers})))
+        selected-viewer)
+      (loop [vs viewers]
+        (if-let [{:as matching-viewer :keys [pred]} (first vs)]
+          (if (and (ifn? pred) (pred v))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/nextjournal/clerk/viewer.cljc [hunk 1]
```diff
@@ -352,42 +350,38 @@
...
-         (if-let [{:as matching-viewer :keys [pred]} (first vs)]
-           (if (and (ifn? pred) (pred v))
-             (apply-viewer viewers matching-viewer v (extract-view-opts x))
-             (recur (rest vs)))
-           (throw (ex-info (str "cannot find matchting viewer for `" (pr-str v) "`") {:viewers viewers :x x :v v}))))))))
+   (let [{:as viewer :keys [render-fn transform-fn]} (lookup-viewer-for x viewers)
+         opts (when (wrapped-value? x)
+                (select-keys x [:nextjournal/width]))
+         v (if transform-fn
+             (-> (value x) value transform-fn)
```
zampino, line 379: The repeated call to `value` is intentional here right?

mk:
@philomates I've implemented this stealing your good names and ideas in #137, thank you for it!

PR #131, opened 2022-04-11
Describe notebook using viewer api and use it to implement slideshow mode 🛝

zampino:
Extend the reach of the viewer api to allow for customization of the whole document, starting from the root node. Show how to use it to create a slideshow viewer.


src/nextjournal/clerk/viewer.cljc
```diff
@@ -180,7 +186,160 @@
       (with-viewer :html
         (into (mkup-fn node) (cond text [text] content (map with-md-viewer content)))))))
 
-(declare !viewers)
+;; block/result formatting fns
+#?(:clj
+   (do
```
zampino, line 191: This is pretty hørrible. Moreover, moving [toward clj-only viewer selection](https://github.com/nextjournal/clerk/pull/131/commits/d1e89c154621200d7d90319a34419484dc858fd6), the functions from this ns to be shared with cljs should reduce to just a very few (correct?). Maybe better settle on that before proceeding with this refactoring? Btw @mk I'm looking for prior work around removing viewer-selection from the client, but can't find it.

src/nextjournal/clerk/viewer.cljc
```diff
@@ -617,13 +785,8 @@
                   (instance? clojure.lang.Namespace scope)
                   (var? scope)))
       (swap! !viewers assoc scope viewers)
-      (with-viewer :eval! `'(v/set-viewers! ~(datafy-scope scope) ~viewers)))))
-
```
zampino, line 621: Here I'm dropping registration in the client and removing `:eval!` viewer. `set-viewers!` returns the collection it's been called with.

src/nextjournal/clerk/sci_viewer.cljs
```diff
@@ -188,19 +188,17 @@
...
+                                       (when viewer-name (str "viewer-" (name viewer-name)))
                                        (when inner-viewer-name (str "viewer-" (name inner-viewer-name)))
-                                       (case (or (viewer/width x) (case viewer (:code :code-folded) :wide :prose))
+                                       (case (or (viewer/width x) (case viewer-name (:code :code-folded) :wide :prose))
                                          :wide "w-full max-w-wide"
                                          :full "w-full"
                                          "w-full max-w-prose px-8")]}
-                         (cond-> [inspect x]
-                           blob-id (with-meta {:key blob-id}))])))
+                         y])))
```
zampino, line 201: Here blocks can be inlined as they come, since with the top-level `fetch-all` all wrapped value are inspected. Was surprised that this just works, but didn't see any troubles so far.

src/nextjournal/clerk/viewer.cljc
```diff
@@ -180,7 +182,119 @@
...
+(defn fn->str [f]
+  (let [pr-rep (pr-str f)
+        f-name (subs pr-rep (count "#function[") (- (count pr-rep) 1))]
+    f-name))
+
+#_(fn->str (fn []))
+#_(fn->str +)
+
+;; TODO: consider removing this and rely only on viewers
+(defn make-readable [x]
```
zampino, line 199: Maybe we can just drop this altogether, as now describe should make both functions and vars readable by processing such types with their relevant viewers. But I'd propose we change this in a follow up step.

We might still need the meta removal step.

src/nextjournal/clerk/viewer.cljc
```diff
@@ -16,7 +16,10 @@
                    (javax.imageio ImageIO)
                    (java.util Base64))))
 
-(defrecord ViewerEval [form])
+(defrecord ViewerEval [form]
+  #?@(:cljs [IFn
+             (-invoke [_this x] ((eval form) x))
+             (-invoke [_this x y] ((eval form) x y))]))
```
zampino, line 22: This is sub-optimal but allows to treat `ViewerEval` instances as valid hiccup/reagent tags. It's needed when calling `viewer/describe` (and hence `viewer/inspect-leafs`) in a cljs context. Maybe we can find a more explicit approach or rename `ViewerEval` which so far should only be called with `v/inspect` as its form.

If we don't implement `IFn` we'll get: 

```
Error
Assert failed: Invalid Hiccup form: [#nextjournal.clerk.viewer.ViewerEval{:form v/inspect} {:path [], :nextjournal/value ["h3" {:id "Hello%20Markdown"} [#nextjournal.clerk.viewer.ViewerEval{:form v/inspect} {:path [], :nextjournal/value [:span "Hello Markdown"], :nextjournal/viewer {:name :html, :render-fn #nextjournal.clerk.viewer.ViewerFn{:form v/html, :f nextjournal.clerk.sci_viewer.html_viewer}}}]], :nextjournal/viewer {:name :html, :render-fn #nextjournal.clerk.viewer.ViewerFn{:form v/html, :f nextjournal.clerk.sci_viewer.html_viewer}}}]
 (in nextjournal.clerk.sci_viewer.inspect)
(valid-tag? tag)
```

Note that we need to prepend a describe to all wrapped values with viewers which don't have a render-fn. This is the case of the markdown viewer: we'll need expressions like

 ```clojure
[inspect (viewer/describe (md "# Markdown"))]
```

for it to render correctly.

File: src/nextjournal/clerk/static_app.cljs [hunk 6]
```diff
@@ -24,25 +24,24 @@
...
+    [:span
+     "Generated with "
+     [:a.hover:text-indigo-500.dark:hover:text-white.font-medium.border-b.border-dotted.border-gray-300
+      {:href "https://github.com/nextjournal/clerk"} "Clerk"]
+     (when (and url sha (contains? url->path path))
+       [:<>
+        " from "
+        [:a.hover:text-indigo-500.dark:hover:text-white.font-medium.border-b.border-dotted.border-gray-300
+         {:href (str url "/blob/" sha "/" (url->path path))} (url->path path) "@" [:span.tabular-nums (subs sha 0 7)]]])]]
+   [sci-viewer/root]])
```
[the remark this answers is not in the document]
zampino: Decided to restore the code as it was before, just handled the slideshow notebook case not to crash static app.

PR #140, opened 2022-04-28
Flatter DOM structure

joe-loco:


mk:
Unsure we still want this and if we do, there's probably an easier way to do this by walking the datastructure and merging them before sending them over the wire.

PR #143, opened 2022-05-02
Upgrade babashka.fs to fix static build on windows

### LINKED ISSUE #141: `clerk/expand-paths` not working on windows
Thank you for creating clerk!

The static site generation fails in windows, tracking down the issue with martin resolved in finding this: 
```clojure
(clerk/expand-paths ["notebooks\\rule_30.clj"])
;; => []
```

Martin remarked:
"wondering if the `fs/glob` is a problem https://github.com/nextjournal/clerk/blob/main/src/nextjournal/clerk.clj#L457
[clerk.clj](https://github.com/nextjournal/clerk/blob/main/src/nextjournal/clerk.clj)"

borkdude:
Fixes #141.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,7 +93,7 @@ jobs:
             ${{ runner.os }}-maven-
 
       - name: 🧪 Run tests
-        run: bb clojure -X:test
+        run: bb test:clj
 
   static-build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


borkdude:
Perhaps we need an additional test for this?

mk:
Yep, a test for expand-paths would be good. I didn’t want to write it without knowing if the implementation would normalize the fs separators.

borkdude:
@mk Added a test and verified that the test fails with fs 0.1.4, but works with fs 0.1.5

mk:
@borkdude excellent, thank you!

PR #149, opened 2022-05-16
Support ShadowCLJS nREPL dev flow on IntelliJ/Cursive

zampino:
This facilitates Cursive folks in the development of Clerk's CLJS layer. The present dev environment is geared toward Cider users only.

1. use Shadow provided nREPL `bb dev :shadow-nrepl? true`
2. skip loading Cider middleware on stock nREPL `bb dev :load-cider-middleware? false`


mk:
Don't think this extra complexity is worth it. If we do want/need the CLJS repl we should just always let shadow start it.

PR #150, opened 2022-05-16
Fix parsing and printing RegEx patterns

zampino:
These changes address two issues around Clerk handling of Regular Expressions: 

1. **Parsing**: rewrite-clj node tag missing
2. **Printing**:

We currently stack overflow the browser when a notebook tries to print a value which:

- doesn't have a dedicated matching viewer but falls back to `:read+inspect`
- is pr-str-able in cljs and "reads-back-to-itself" via edamame

This is because the render-fn of `:read+inspect` calls `describe` in the client (via `inspect-paginated`)
```clj
{:name :read+inspect
 :render-fn '(fn [x] (v/html [v/inspect-paginated 
                               (try (v/read-string x)
                                    (catch js/Error _e
                                      (v/with-viewer v/unreadable-edn-viewer x)))]))}
```
which, for such values, results in an endless loop of `pr-str -> read-string -> describe -> pr-str ...`. This is the case for Regex pattern instances but might be true for other types as well.

The issue doesn't apply to values read (or described) as tagged values, as most of the test cases in the printing notebook (instants, UUIDs, etc.).


mk:
I've pushed a simpler version of this with cbfefc5d37efdf00f1f77a696e5fddb057e495e0 to #152.

PR #160, opened 2022-06-01
Add goog.i18n.NumberFormat

borkdude:


mk:
We no longer need this I think.

PR #172, opened 2022-06-13
Markdown Fixes (Inline Images + loose lists)

zampino:
- [x] Bumps Markdown (from new home)
- [x] awaits merge of https://github.com/nextjournal/viewers/pull/70
- [x] Fixes [tight vs. loose markdown lists](https://snapshots.nextjournal.com/markdown/build/1931cd2878b5b82430c71804775c553281c448a2/index.html#/notebooks/tight_lists.clj)
- [x] Fixes breaking line for images in the same paragraph: 

CleanShot 2022-06-13 at 10 18 49@2x

as in this [build](https://snapshots.nextjournal.com/markdown/build/18fe70da1e4054cfd5cd7c9d933ef687ac317909/index.html#/README.md).


zampino:
There's still an issue with image margins (see [notebook](https://snapshots.nextjournal.com/clerk/build/53d7c2a046030cbf0308be0b93b17bf4537cfbb5/index.html#/notebooks/viewers/custom_markdown.clj)):
CleanShot 2022-06-13 at 13 59 59@2x

Maybe this is because we regard images as being blocks. 

Our markdown structure can already distinguish different scenarios where image is the only child (not wrapped)
```
<div.notebook><img></div>
```
from those where an image has siblings (wrapped in a `<p>`)
```
<div.notebook><p>text <img> text</p></div>
```
do we want to this to be reflected in the CSS?

zampino:
This would circumvent the selector that adds the 2rem top margin, but is nonsense... how could I fix the prose voodoo selector thing?

```
diff --git a/src/nextjournal/clerk/viewer.cljc b/src/nextjournal/clerk/viewer.cljc
index 0e90a2d..cd84382 100644
--- a/src/nextjournal/clerk/viewer.cljc
+++ b/src/nextjournal/clerk/viewer.cljc
@@ -446,7 +446,7 @@
     :transform-fn (into-markup
                    (fn [{:as node :keys [heading-level]}]
                      [(str "h" heading-level) {:id (uri.normalize/normalize-fragment (md.transform/->text node))}]))}
-   {:name :nextjournal.markdown/image :transform-fn #(with-viewer :html [:img.inline (-> % ->value :attrs)])}
+   {:name :nextjournal.markdown/image :transform-fn #(with-viewer :html [:div.inline-block [:img.inline (-> % ->value :attrs)]])}
    {:name :nextjournal.markdown/blockquote :transform-fn (into-markup [:blockquote])}
 ```

PR #173, opened 2022-06-13
Preserve `*ns*` during analysis and eval

mk:


resources/viewer-js-hash
```diff
@@ -1 +1 @@
-MZAFz7cCjz5iXe9gpb2mNybfmhh
\ No newline at end of file
+2QomkaERveEgZwKtSGc6f95nabbt
```
zampino, line 1: This should not have changed given the file touched, I believe.

PR #175, opened 2022-06-14
Fix #171: support loading render fns from .cljs source

### LINKED ISSUE #171: Provide a way to load a viewer function from a file
In PR #155 which provides a SCI nREPL implementation, there is a hack to load source from a file and then evaluate within SCI + resolving a function after that. When we move to async SCI evaluation (e.g. to load stuff from CDNs using `require`) such has [here](https://nextjournal.com/try/borkdude/confetti), this trick doesn't work anymore. Instead, we should provide a nice API for loading viewer functions from files. E.g.:

`:render-fn {:path "..." :fn foo/bar}` or `:render-fn {:resource "..." :fn foo/bar}`.

borkdude:


notebooks/render_fn.clj
```diff
@@ -0,0 +1,14 @@
+(require '[nextjournal.clerk :as clerk])
+
+#_(clerk/clear-cache!)
```
borkdude, line 3: @mk This should not be necessary anymore after you implement content-hashing on atom contents

src/nextjournal/clerk.clj
```diff
@@ -508,6 +508,18 @@
         {state :result duration :time-ms} (time-ms (write-static-app! opts state))]
     (report-fn {:stage :finished :state state :duration duration :total-duration (elapsed-ms start)})))
 
+(defn render-fn* [sym]
```
borkdude, line 511: This looks up the render fn source for evaluation on the client

src/nextjournal/clerk/viewer.cljc
```diff
@@ -1,21 +1,23 @@
 (ns nextjournal.clerk.viewer
```
borkdude, line 1: Sorry, this is just me using lsp-organize-import, hope you don't mind

src/nextjournal/clerk.clj
```diff
@@ -508,6 +508,18 @@
...
+  (let [ns (namespace sym)
+        file (str (munge ns) ".cljs")]
```
zampino, line 513: I'm trying this with a symbol `scratch.freehand/render-fn` (file lives in `notebooks/scratch/freehand.cljs`) munge alone doesn't convert `.` in `/` right? Or do we make some assumptions on the location of the cljs files?

borkdude: Good catch! We have to convert `"."` to `"/"`, I will push that.

mk:
I just merged https://github.com/nextjournal/clerk/pull/187 into `main`. Hope this works together as we would expect. We could also use `:nextjournal.clerk/hash-fn` if we prefer that over implementing `IDeref`:
https://github.com/nextjournal/clerk/blob/3c27ad52da8173ceb92d411af0957c29a75faf4b/notebooks/valuehash.clj#L14-L19

borkdude:
Thanks, I'll take a look if I can make .cljs source work with the `hash-fn` instead of deref.

PR #180, opened 2022-06-20
Fix #147: inline asset map in mvn release

### LINKED ISSUE #147: Allow clerk to function behind a proxy / bundle dependencies in a release
https://github.com/nextjournal/clerk/issues/30

This is linked to the above issue in my work environment. However, in my work environment while we can download maven dependencies through a certain system, we will not be able to access links like "https://storage.googleapis.com/nextjournal-cas-eu" 

Which seems to be a necessity as soon as `(require '[nextjournal.clerk :as clerk])` is ran

borkdude:


mk:
Looks good, thanks! Wondering if it could be simpler to look up the asset map from an `io/resource` and fall back to the slurp in case it doesn't exists? This would maybe also make testing easier maybe. But fine to merge it as-is if you prefer this.

PR #181, opened 2022-06-20
Preserve filename in cas file

borkdude:


bb/viewer_resources_hashing.clj
```diff
@@ -51,8 +51,10 @@
 (defn lookup-url [lookup-hash]
   (str gs-bucket "/lookup/" lookup-hash))
 
-(defn cas-link [hash]
-  (str base-url "/data/" hash))
+(defn asset-name [hash suffix]
+  (str "/data/" hash
```
mk, line 55: Let's maybe keep the things in `/data` content-addressed and move this to `/assets`.

mk (approved):
Slight preference for `/assets` in the name but good to go either way.

PR #182, opened 2022-06-20
Fix bug in sci-viewer/tagged-value

borkdude:


src/nextjournal/clerk/sci_viewer.cljs
```diff
@@ -514,7 +514,7 @@
              trace)]]]]))
 
 (defn tagged-value
-  ([tag value] ({:space? true} tagged-value tag value))
+  ([tag value] (tagged-value {:space? true} tag value))
```
borkdude, line 517: Found this bug using clj-kondo...

PR #183, opened 2022-06-20
Automate release

borkdude:
- Auto-publish to CI when tag is pushed
- Expose jar as artifact


mk:
One thing that would be also nice is setting the `:tag` in `:scm` in `b/write-pom`. Currently the backlinks from clojars https://clojars.org/io.github.nextjournal/clerk/versions/0.8.470 just point at HEAD:
> [mkvlr](https://clojars.org/users/mkvlr) on Jun 20, 2022 with [this git tree](http://github.com/nextjournal/clerk/tree/HEAD)

borkdude:
@mk Sure.

Can you configure these secrets?

CLOJARS_USERNAME
CLOJARS_PASSWORD

The password can be a token, not your personal clojars password.

mk:
@borkdude I've setup `CLOJARS_PASSWORD_MKVLR` as an organization secret (so we can use it in other repos as well, doing the same automation for https://github.com/nextjournal/markdown might be a good next step). I guess we can leave `CLOJARS_USERNAME=mkvlr` in plain text in the workflow or let me know if this should also be a secret and I'll set it up.

borkdude:
@mk Perfect

build.clj
```diff
@@ -27,7 +27,8 @@
                 :basis basis
                 :scm {:url "http://github.com/nextjournal/clerk"
                       :connection "scm:git:git://github.com/nextjournal/clerk.git"
-                      :developerConnection "scm:git:ssh://git@github.com/nextjournal/clerk.git"}
+                      :developerConnection "scm:git:ssh://git@github.com/nextjournal/clerk.git"
+                      :tag (str "v" version)}
```
zampino, line 31: Nice, this will fix the backlink from clojars!

PR #191, opened 2022-07-06
Bump clojure-mode & inspect SCI Vars

zampino:
* bumps viewers repo with upgraded `io.github.nextjournal/clojure-mode` that upgrades to CodeMirror 6.0 release
* bump `io.github.nextjournal/markdown`
* support inspecting sci vars


zampino:
I noticed that with just `bb dev` shadow doesn't update `package.json` according to `deps.cljs` in classpath (In other projects I believe it does). I had to explicitely run `clj -M:demo:sci:dev npm-deps`.

zampino:
I took the opportunity to fix devcards app as it was looking somehow odd stylewise (see below). I propose to sync tailwind stylesheet [when dev environment boots](https://github.com/nextjournal/clerk/pull/191/commits/5c9b2c9970e1d538ba0d2aef89907f190af81bd4) with `bb dev`. Any better idea? I can also push the 2 commits out to somewhere else.

CleanShot 2022-07-06 at 18 10 21@2x

PR #195, opened 2022-08-03
Unify cljs and clj viewer API

zampino:
Allow projects using Clerk's sci-viewer context to use the same API when evaluating clojurescript with SCI. Also do a naming pass:
* `inspect-paginated` → `inspect`
* `inspect` → `inspect-presented`

Used in this [notebook](https://snapshots.nextjournal.com/clojure-mode/build/f8e6ef396b71150ae88715c5aecabc296f4838a8/index.html#preview-it).

CleanShot 2022-08-03 at 15 45 58@2x


src/nextjournal/clerk/sci_viewer.cljs
```diff
@@ -1139,7 +1136,20 @@ black")}]))}
 
    'doc-url doc-url
    'url-for url-for
-   'read-string read-string})
+   'read-string read-string
+
+   ;; clerk viewer API
+   'code code
+   'col col
+   'html html-render
```
zampino, line 1144: For symmetry,  I'd like to get `html` as exported from viewer ns, by setting the html-viewer's render-fn to `html-render` there. With that I'm running in all kind of endless recursion troubles, hard to un-🧶.

All calls to `(html x)` in sci-viewer ns seems to produce some unnecessary (?) indirection when going through the  `render-with-viewer` function.

mk (commented):
This is a good first step. Down the road I think it makes sense to further unify the api so we have a single namespace that folks can use it from regular cljs and split up the sci part to a separate config.

PR #196, opened 2022-08-03
Support simple maps in table viewer

Heliosmaster:
With this PR I propose to add support for simple maps in the table viewer. Note that this behavior is already possible by just calling `seq` on the map (which is what we're doing anyway in the implementation).

But I think that a table representation of a map (with keys and values) makes sense enough for the API to be explicitly supported.

CleanShot 2022-08-03 at 19 34 40@2x

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -213,7 +213,7 @@
   Performs normalization on the data, supporting:
   * seqs of maps
   * maps of seqs
-  * seqs of seqs
+  * seqs of seqs (including maps)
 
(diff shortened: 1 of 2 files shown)


mk:
Clerk already supports map of seqs in a table using the map keys as the header row. For displaying a plain map in cols like you posted explicitly converting it into a `seq` is perfectly fine I think, or use `update-vals` if you like to have it with the keys as the header row.

Screen Shot 2022-08-03 at 19 01 23

For displaying a plain map we can maybe think of providing a property list viewer but I don't think it belongs into the table viewer.

PR #203, opened 2022-08-23
Cut down on SCI viewer dependencies

zampino:
A first step toward reducing the deps of the sci-viewer ns.

- [x] Drops re-frame dep
- [x] Drops devcards dep
- [x] Replaces devcards by means of [a clojure notebook](https://snapshots.nextjournal.com/clerk/build/e2ef8e58ae3a5d3724b92e0b572466b25e3718a9/index.html#/notebooks/cards.clj) with a custom viewer for inspecting (cl)js expressions. It is _less_ general than devcards in that it can only evaluate code in the context of `n.c.sci-viewer/!sci-ctx`.
- [x] Fix analyzer issues in static build for cards notebook

Will work towards dropping the dependency on the nextjournal viewers in a follow-up.

In addition, this improves the inspection of JS objects: prior to this we were falling back to the fallback viewer (pr-str + read + inspect). In particular it now doesn't fail to display nested objects and we can e.g. inspect `js/window`.


src/nextjournal/clerk/sci_viewer.cljs
```diff
@@ -1165,7 +715,3 @@ black")}]))}
   (sci/eval-form @!sci-ctx f))
 
 (set! *eval* eval-form)
-
-(swap! viewer/!viewers (fn [viewers]
-                         (-> (into {} (map (juxt key (comp #(into [] (map viewer/process-render-fn) %)  val))) viewers)
-                             (update :root concat js-viewers))))
```
zampino, line 1171: `js-viewers` were placed at key `:root` and hence unused, and I'm not sure we need to process render fns as before at all.

zampino: Render fns are "processed" as a result of the presentation of a value.

src/nextjournal/clerk/analyzer.clj
```diff
@@ -76,8 +76,12 @@
...
   ([form] (analyze-form {} form))
   ([bindings form]
-   (binding [config/*in-clerk* true]
-     (ana-jvm/analyze form (ana-jvm/empty-env) {:bindings bindings}))))
+   (try
+     (binding [config/*in-clerk* true]
+       (ana-jvm/analyze form (ana-jvm/empty-env) {:bindings bindings}))
+     (catch Throwable e
+       (throw (ex-info (str "cannot analyze form " form ": " (ex-message e))
+                       {:form form :bindings bindings} e))))))
```
mk, line 84: Can you undo this change? I'd like to have this separately and I'm unsure this is the right place for the catch, want to surface more context (the file etc).

src/nextjournal/clerk/viewer.cljc
```diff
@@ -1254,3 +1285,8 @@
                  (v/html (into [:div.border-l-2.border-slate-300.pl-4
                                 [:div.uppercase.tracking-wider.text-xs.font-sans.text-slate-500.mt-4.mb-2 "Examples"]]
                                (v/inspect-children opts) examples)))})
+;; cards
+(def card-viewer
+  {:transform-fn (comp mark-presented (update-val ->viewer-eval))
+   :render-fn '(fn [data]
+                 (if (v/valid-react-element? data) data (v/html [v/inspect data])))})
```
mk, line 1292: Can we remove this?

src/nextjournal/clerk.clj
```diff
@@ -306,6 +306,10 @@
     `(clerk/with-viewer v/examples-viewer
        (mapv (fn [form# val#] {:form form# :val val#}) ~(mapv (fn [x#] `'~x#) body) ~(vec body)))))
 
+(defmacro card
+  "Evaluates and inspects `body` in SCI context for cljs consumption."
+  [body]
+  `(clerk/with-viewer v/card-viewer '~body))
```
mk, line 312: I'd rather not have this in the public api. Can you inline it in the notebook for now, even if it means we need to manually evaluate it?

PR #204, opened 2022-08-25
Make Parser cljc

zampino:
* makes parser namespace cljs compatible
* add card


mk (approved):
Build size impact is 4k so less than 0.1%, merging this.

PR #216, opened 2022-09-22
Better Error Reporting during Static Build

zampino:


src/nextjournal/clerk/builder.clj
```diff
@@ -139,18 +139,26 @@
...
+        state (mapv (fn [{:as doc :keys [file]}]
                       (report-fn {:stage :building :doc doc})
                       (let [{doc+viewer :result duration :time-ms} (eval/time-ms
-                                                                    (let [doc (eval/eval-analyzed-doc doc)]
-                                                                      (assoc doc :viewer (view/doc->viewer (assoc opts :inline-results? true) doc))))]
+                                                                    (try
+                                                                      (let [doc (eval/eval-analyzed-doc doc)]
+                                                                        (assoc doc :viewer (view/doc->viewer (assoc opts :inline-results? true) doc)))
+                                                                      (catch Exception e
+                                                                        (throw (ex-info (str "cannot build notebook: " file) {:file file} e)))))]
```
zampino, line 161: This might be not strictly necessary, given the printing of the notebook path via `report-fn`, but might be more easily discoverable and closer to the exception. It also adds the file as ex-data.

zampino:
Let's revisit this by improving reports closer to the analyzer, not just for the static build.

PR #224, opened 2022-10-03
Fix results with `*print-length/depth*` being set

russmatney:
If `*print-length*` is set (e.g. in a cider config), it can break edamame
parsing in the frontend - collections will be limited to whatever that
length is, and an extra `...` will be added to the printed edn, causing
an odd number of forms to show up in maps (making them unparsable).

---

I just got to the bottom of a bug that I thought was in Clerk, but it turned out to
be my own config setting `*print-length*`.

I've been consistently seeing errors like:

```
The map literal starting with [0 2 1 2 1 1 0 1 0] contains 101 form(s). Map literals must contain an even number of forms.
```

The `101` here was a clue - I eventually realized it was the `*print-length*`, which I
set to `50` in my cider config to reduce lag when evaling large values:

``` emacs-lisp
(setq cider-repl-init-code (append cider-repl-init-code '("(set! *print-length* 50)")))
```

For now I've updated my clerk-show and other clerk bindings to unset *print-length*:

``` emacs-lisp
(defun clerk-show ()
  (interactive)
  (save-buffer)
  (let
      ((filename (buffer-file-name)))
    (when filename
      (cider-interactive-eval
       (concat "
(with-bindings {#'*print-length* nil}
  (nextjournal.clerk/show! \"" filename "\"))")))))
```

This PR binds `*print-length*` to `nil` in `clerk.viewer/->edn`, which resolves the
issue on my end (i.e. no binding change necessary). I'm not sure if there are
other locations this needs to be set, but I haven't experienced it elsewhere
after making this change.


mk:
@russmatney good catch, thank you!

PR #236, opened 2022-10-21
Keep processing even if one of the notebooks fail

pfeodrippe:
Adds `:best-effort` flag.

Let me know if you think this is a good idea.

In my case, one notebook was failing because Clerk was complaining about not seeing the `map->MyRecord` var (or something similar) from a record `MyRecord` in the namespace and the other was that a macro that used an aliased require (`log/...`) was considered as non-existent by Clerk (in run time, it works fine).

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -425,6 +425,7 @@
   - `:bundle`    - if true results in a single self-contained html file including inlined images
   - `:browse`    - if true will open browser with the built file on success
   - `:dashboard` - if true will start a server and show a rich build report in the browser (use with `:bundle` to open browser)
+  - `:best-effort` - if true will keep processing other notebooks even in case of some failing
   - `:out-path`  - a relative path to a folder to contain the static pages (defaults to `\"public/build\"`)
   - `:git/sha`, `:git/url` - when both present, each page displays a link to `(str url \"blob\" sha path-to-notebook)`
   "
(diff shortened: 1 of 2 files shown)


mk:
Hi @pfeodrippe,

I'd love to see an issue and small repro about the Clerk error you've seen. When showing a namespace, Clerk tries to be more strict than the REPL and throws if you're referring to vars that no longer exist in the file after you've removed them.

I'm aware of one macro running into an issue with this and it sounds like you've got another one. Would like to get to the bottom of this.

Regarding the change here, I think I'd prefer the keep build failing loudly if any notebook fails. Your change only affects analysis but there's also the parsing & eval phases where a single notebook can fail. I think we'd have to consider all phases for such an `:continue-on-error?` flag but I'd like to wait a bit until this code settles down more and see if we really need it.

pfeodrippe:
> Hi @pfeodrippe,
> 
> I'd love to see an issue and small repro about the Clerk error you've seen. When showing a namespace, Clerk tries to be more strict than the REPL and throws if you're referring to vars that no longer exist in the file after you've removed them.
> 
> I'm aware of one macro running into an issue with this and it sounds like you've got another one. Would like to get to the bottom of this.
> 
> Regarding the change here, I think I'd prefer the keep build failing loudly if any notebook fails. Your change only affects analysis but there's also the parsing & eval phases where a single notebook can fail. I think we'd have to consider all phases for such an `:continue-on-error?` flag but I'd like to wait a bit until this code settles down more and see if we really need it.

Sure, will open them tonight, thanks :)

mk:
Thanks @pfeodrippe! I'm closing this, we might want to revisit it once the builder code has stabilised.

pfeodrippe:
> Thanks @pfeodrippe! I'm closing this, we might want to revisit it once the builder code has stabilised.

Sure. thanks!

PR #240, opened 2022-10-22
try unkeyed-list

mhuebert:
Maybe something changed in Reagent (or w/ functional components), I thought previously that `(into [:<>] ...)` would have the same effect as this.

Does this address the key issues @mk? I am able to still re-render plotly without remounts, and I don't see key errors, but have not tested any other cases.


mk:
@mhuebert the duplicate key issues are fixed by only removing `^{:key (str viewer (:idx opts))}` from `inspect-presented`, but that brings back the `Rendered fewer hooks than expected. This may be caused by an accidental early return statement.` errors we've already seen before https://github.com/nextjournal/clerk/pull/231#issuecomment-1279768408.

mhuebert:
Ah good point.

PR #243, opened 2022-10-24
Add Open Graph Metadata to static Build

zampino:
* default we will use the last suitable image (sizewise) in the notebook, if there is none, we take the screenshot of the whole notebook
* folks need to be able to flag a result so they can override this default
* folks can also specify a url which contains the image
* do we allow overriding the opengraph text part as well? How do we autogenerate it?
* what about the favicon
* do we set the title in the doc?
* target unbundled case first

https://nextjournal.slack.com/archives/C022KRVLAC9/p1666611149563889

https://ogp.me/#metadata


zampino:
* proposed API for specifying OG details, according to https://ogp.me/#metadata

> To turn your web pages into graph objects, you need to add basic metadata to your page. We've based the initial version of the protocol on RDFa which means that you'll place additional <meta> tags in the <head> of your web page. The four required properties for every page are:
> * og:title - The title of your object as it should appear within the graph, e.g., "The Rock".
> * og:type - The [type](https://ogp.me/#types) of your object, e.g., "video.movie". Depending on the type you specify, other properties may also be required.
> * og:image - An image URL which should represent your object within the graph.
> * og:url - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "https://www.imdb.com/title/tt0117500/".

which could be matched by namespace meta like:

https://github.com/nextjournal/clerk/blob/b972be682e9faf947cd6264c72f6695146acea58/notebooks/open_graph.clj#L2-L17

missing keys could be filled in by us.

Possible ns keys:
* `:nextjournal.clerk/open-graph`
* `:nextjournal.clerk/og`

Scalar values, a single key for each property

* `:nextjournal.clerk/og-url`

joe-loco:
Some thoughts:

I think I would keep this analogous to the `visibility` options, i.e.

On the ns level:

```cli
(ns example
  {:nextjournal.clerk/visibility {:code :fold :result :hide}
   :nextjournal.clerk/open-graph {:title "Example"
                                  :type "article:clerk"
                                  :image "https://example.dev/og-image.png"
                                  :description "A Clerk notebook explaining how to use Open Graph tags with your notebook."}})
```

```cli
{::clerk/visibility {:code :fold :result :hide}
 ::clerk/open-graph {:title "Example"
                     :type "article:clerk"
                     :image "https://example.dev/og-image.png"
                     :description "A Clerk notebook explaining how to use Open Graph tags with your notebook."}}
```
All keys in `::clerk/open-graph` map to the OG spec tags minus the `og-` prefix and all values should follow the spec. We should set up some values by default:

* image (the result image that best fits into a 1200x630 (or 1.91:1 aspect ratio) rectangle. If none is present the notebook preview will be used.
* `title` the first `#` level one heading that’s used in the document. This should also be used for `<title>`.
* `type` wil be `article:clerk`
* `description` will be the first two sentences of the document. I’m not sure if we should use headings here as well or just paragraphs. I’m also not sure if this should be set by default at all. It might be that we pick something that doesn't make sense.

To pick a result as image, I propose following the `visibility` approach as well, i.e.

```clj
^{::clerk/visibility {:code :fold :result :hide}
  ::clerk/open-graph {:image :result}
```
The `:result` key follows how we use `:hide` or `:fold` for `visibility` and keeps the possibility open that we use other values than `:result` in the future (as opposed to a boolean). Further, it clearly states what is used as image (the result — again, as opposed to a boolean).

Keeping this in line with `visibility` allows us to re-use already established ergonomics and (maybe?) also use the existing mechanism that we have in place for `visibility`.

joe-loco:
Perhaps related to this: It really always irks me that I only get the OG preview once I use some article in the actual service I want to post it to. So having some form of preview would be super helpful.

zampino:
How should `build-static-app!` enter the play?
1. If some og meta are present or inferred, they should be added in the static page view regardless of screenshots
2. There should be an option to trigger the creation of screenshots. 
   In that case, the user need to provide the URL at which they'll host the static page as we cannot use relative URL.

In case of 1. we don't need to wait the actual page creation for populating ogs in the view. In case of 2. we need to first generate the page and then add the og image url.

zampino:
> the first # level one heading that’s used in the document. This should also be used for <title>.

A parsed document has a `:title` key.

> description — will be the first two sentences of the document. I’m not sure if we should use headings here as well or just use paragraphs

We could pick the text of the first node of kind paragraph as parsed by markdown.

zampino:
At this point, this already allows for having card image URLs configured explicitly

https://github.com/nextjournal/clerk/blob/2b79bba4fe1fb296ae00d49434dec36a626749ae/book.clj#L3-L7

CleanShot 2022-10-26 at 11 20 59@2xCleanShot 2022-10-26 at 11 17 49@2x

Shall we add the snapshotting feature in a follow up?

PR #246, opened 2022-10-26
Revive Tailwind CSS Compilation

zampino:
## Workflow

* opt-in during build static app or (bb task?)
* run tailwind command against static pages
  * input: stylesheets from resources
  * config: tw config from resources
  * content: html pages + js (build js locally or copy from CAS)
* check bundle vs unbundled case
* check with custom viewers producing hiccup/html
* ~upload purged CSS to CAS (can we use dejavu  ?)~ copy to out folder
* tailwind play cdn link need to be replaced with stylesheet link  pointing at the ~CAS~ URL.

## API

If it's meant to be part of `build-static-app!` we could start by passing an `:optimize-css-fn` option. The function is responsible for the whole workflow above except for the string replacement, and returns a stylesheet URL. We could handle the replacement on our side.

## Current Config

```
{
   darkMode: \"class\",
   content: [\"./public/build/index.html\", \"./public/build/**/*.html\", \"./build/viewer.js\"],
   safelist: ['dark'],
   theme: {
     extend: {},
     fontFamily: {
       sans: [\"Fira Sans\", \"-apple-system\", \"BlinkMacSystemFont\", \"sans-serif\"],
       serif: [\"PT Serif\", \"serif\"],
       mono: [\"Fira Mono\", \"monospace\"]
     }
   },
   variants: {
     extend: {},
   },
   plugins: [require(\"@tailwindcss/typography\")],
 }
```

## Prior Art

clerk

https://github.com/nextjournal/clerk/blob/914c31026ef8285b5af987277f70f01722cfab61/src/nextjournal/clerk.clj#L357-L366

nextjournal

https://github.com/nextjournal/nextjournal/blob/5a5f01ccb49a70885e272ef65db1589442345cad/ops/src/nextjournal/ops/build.clj#L25-L46


deps.edn
```diff
@@ -25,8 +25,10 @@
                                             org.babashka/cli {:mvn/version "0.5.40"}}
                                :extra-paths ["notebooks"]
                                :exec-fn nextjournal.clerk/build!
-                               :exec-args {:paths-fn nextjournal.clerk.builder/clerk-docs}
+                               :exec-args {:paths-fn nextjournal.clerk.builder/clerk-docs
+                                           :compile-css? true}
```
mk, line 29: Drop the ? here please, (like we did for bundle) we normalize this early

src/nextjournal/clerk/builder.clj
```diff
@@ -78,6 +81,7 @@
                                  (str "Errored in " duration ". ❌\n")
                                  (str "Done in " duration ". ✅\n"))
       :building (str "🔨 Building \"" (:file doc) "\"… ")
+      :compiling-css "🎨 Optimizing CSS… "
```
mk, line 84: Compiling Tailwind CSS…

src/nextjournal/clerk/builder.clj
```diff
@@ -183,8 +187,30 @@
       (expand-paths {:paths-fn `my-paths}))
 #_(expand-paths {:paths ["notebooks/viewers**"]})
 
+(defn compile-css!
+  "Compiles a minimal tailwind css stylesheet with only the used styles included, replaces the generated stylesheet link in html pages."
+  [{:keys [out-path]} docs]
+  (assert (and (= 0 (:exit (sh "which" "npx"))) (= 0 (:exit (sh "npx" "tailwindcss"))))
+          "Clerk's CSS optimizaiton failed: node and tailwind need to be installed. Please run `npm install -D tailwindcss @tailwindcss/typography` and retry.")
+  (spit "tailwind.config.cjs" (slurp (io/resource "stylesheets/tailwind.config.js")))
```
mk, line 195: Why .cjs and not .js?

src/nextjournal/clerk/builder.clj
```diff
@@ -183,8 +187,30 @@
...
+    (spit (let [path (fs/path "build" (str/replace file #"\.(cljc?|md)$" ".edn"))]
+            (fs/create-dirs (fs/parent path))
+            (str path))
+          (pr-str viewer)))
+  (sh "npx" "tailwindcss"
+      "--input" "input.css"
+      "--config" "tailwind.config.cjs"
+      "--output" (str (fs/path out-path "viewer.css"))
+      "--minify")
+  (swap! config/!resource->url assoc "/css/viewer.css" "viewer.css"))
```
mk, line 209: We should content address the css (like we do for the images) so the file name should be sha512.css

src/nextjournal/clerk/view.clj
```diff
@@ -3,31 +3,44 @@
...
 
-
 #_(doc->viewer (nextjournal.clerk/eval-file "notebooks/hello.clj"))
 #_(nextjournal.clerk/show! "notebooks/test.clj")
 #_(nextjournal.clerk/show! "notebooks/visibility.clj")
 
-(defn include-viewer-css []
+(defn relative? [url]
+  (and (not (.isAbsolute (URI. url)))
+       (not (str/starts-with? url "/"))))
```
mk, line 21: Why are both of these checks needed?

src/nextjournal/clerk/builder.clj
zampino: because of type: module in the package.json. All .js files are treated like ES modules unless they have a .cjs extension.

src/nextjournal/clerk/view.clj
zampino: Because `(= false (.isAbsolute (URI. "/viewer.css")))`

src/nextjournal/clerk/builder.clj
zampino: Oh that was with yarn, npx seems happy either way

PR #257, opened 2022-11-02
Extract hooks to own ns, inline code viewer and let it use hooks

zampino:
* inlines code viewer in render namespace
* uses React hooks to embed codemirror into DOM


src/nextjournal/clerk/render.cljs
```diff
@@ -765,7 +768,13 @@
         default-loading-view))))
 
 (def render-mathjax mathjax/viewer)
-(def render-code code/viewer)
+
+(defn render-code [value]
+  (let [ref (use-ref nil)]
```
zampino, line 773: @mhuebert is `use-effect` right choice here?  Effect cleanup has the codemirror instance in scope in this way. Do you think there's a better way to do this?

zampino:
This can be merged from my side, I think.

PR #258, opened 2022-11-02
Bring back auto-expand and make it opt-in

joe-loco:
Opt-in to auto-expanding results via the `:nextjournal.clerk/auto-expand-results? true` doc setting.


joe-loco:
Please don’t merge this yet. There is still an issue with table viewers when `{:nextjournal.clerk/auto-expand-results? true}`.

PR #259, opened 2022-11-02
Fix content addressing of static app data

zampino:
Fix actual content-addressing of image-blobs and compiled CSS during static build

We previously produced SHAs of nippy-serialized values


src/nextjournal/clerk/viewer.cljc
```diff
@@ -318,18 +320,24 @@
...
 #?(:clj
-   (defn maybe-store-result-as-file [{:as _doc+blob-opts :keys [blob-id file out-path]} {:as result :nextjournal/keys [content-type value]}]
+   ;; TODO: move away from viewer ns
+   (defn store-in-cas! [{:keys [out-path ext]} content]
+     (assert out-path) (assert ext)
+     (let [cas-path (fs/path out-path "_data" (str (multihash/base58 (digest/sha2-512 content)) "." ext))]
+       (fs/create-dirs (fs/parent cas-path))
+       (when-not (fs/exists? cas-path)
+         (Files/write cas-path content (into-array [StandardOpenOption/CREATE])))
+       (str cas-path))))
```
zampino, line 330: Where could I move this? I can't put it in builder since we need to use it in this ns. I could pass the fn down in opts, but sounds more error-prone. A new shared ns?

mk: I think it's fine here for now.

PR #261, opened 2022-11-03
support for :extra-namespaces

mhuebert:
This adds support for an `:extra-namespaces` option for `dev-launcher/start`. 

TODO: support `:extra-namespaces` in release builds


viewer/deps.edn
```diff
@@ -0,0 +1,11 @@
+{:paths ["src"]
```
mk, line 1: Do we need to also add `"../src"` as path here (is that even supported?)? Otherwise we'd be missing the sources. I'd prefer if we could avoid having to move the source files to a different directory.

viewer/deps.edn
```diff
@@ -0,0 +1,11 @@
...
+ :deps {applied-science/js-interop {:mvn/version "0.3.3"}
+        binaryage/devtools {:mvn/version "1.0.3"}
+        org.babashka/sci {:mvn/version "0.5.34"
+                          #_#_#_#_:git/url "https://github.com/babashka/sci"
+                          :git/sha "a766b73f668e8ba59855ce229fad1aee08038922"}
+        reagent/reagent {:mvn/version "1.1.1"}
+        io.github.babashka/sci.configs {:git/sha "b5335b79c237344d003f16358321197a0b0a2ac6"}
+        io.github.nextjournal/viewers {:git/sha "1aaedea7709611cbb393add4c85c7a2dd551260e"}
+        metosin/reitit-frontend {:mvn/version "0.5.15"}
+        thheller/shadow-cljs {:mvn/version "2.20.7"}}}
```
mk, line 11: Unsure if we should add `shadow-cljs` here. 

It's not possible to use an alias from a library, correct?

viewer/src/nextjournal/clerk/shadow-cljs.edn
```diff
@@ -1,14 +1,13 @@
-{:deps {:aliases [:sci]}
- :dev-http {7778 {:roots ["public" "classpath:public"]}}
- :nrepl false
```
mk, line 3: Are we launching two nrepls now?

mk (commented):
Looks good at a glance, will test it more next week.

Think I'd avoid introducing the extra `viewer/src` classpath dir and it might be a little invasive to include `user.clj` on the classpath, let's keep that in `dev/`, same for ssr?

viewer/deps.edn
mhuebert: I believe I got an error or warning when I tried adding a src path outside the directory (that was my first approach)

viewer/src/nextjournal/clerk/shadow-cljs.edn
mhuebert: Just fixed

mk:
One thing I'm only now realising that will need fixing still is the viewer resource hashing, e.g.
https://github.com/nextjournal/clerk/blob/cda88db3944820f6c2800d7ce11a95b27ef5572f/bb/viewer_resources_hashing.clj#L27

The purpose of this is generating a hash under which to look up the built js bundle and only triggering a build when needed. Maybe we can also simplify things by using the git sha as a lookup key and always run the js build, it only takes ~20s on my air.

Also `bb release:js` seems to error for me:
```
bb release:js
yarn install v1.22.19
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.07s.
Checking out: https://github.com/babashka/sci.configs.git at b5335b79c237344d003f16358321197a0b0a2ac6
Execution error (NullPointerException) at clojure.main/main (main.java:40).
Cannot invoke "clojure.lang.IFn.applyTo(clojure.lang.ISeq)" because "f" is null

Full report at:
/var/folders/bn/d52n4q5j577gm8mw0c47s6900000gn/T/clojure-611168137127363888.edn
Error while executing task: build:js
```

mk (commented):
I'll see if I can take a stab at this.

mk:
Getting this failure locally after the merge on `bb dev`:
```
FileNotFoundException: /js/cljs-runtime/cljs_env.js (No such file or directory)
	java.io.FileOutputStream.open0 (FileOutputStream.java:-2)
	java.io.FileOutputStream.open (FileOutputStream.java:293)
	java.io.FileOutputStream.<init> (FileOutputStream.java:235)
	clojure.java.io/fn--11522 (io.clj:230)
	clojure.java.io/fn--11522 (io.clj:230)
	clojure.java.io/fn--11459/G--11428--11466 (io.clj:69)
	clojure.java.io/fn--11496 (io.clj:166)
	clojure.java.io/fn--11496 (io.clj:166)
	clojure.java.io/fn--11472/G--11424--11479 (io.clj:69)
	clojure.java.io/writer (io.clj:119)
	clojure.java.io/writer (io.clj:104)
	clojure.core/apply (core.clj:669)
	clojure.core/spit (core.clj:6956)
	clojure.core/spit (core.clj:6956)
	shadow.build.targets.esm/flush-dev/fn--36704 (esm.clj:415)
	shadow.build.targets.esm/flush-dev (esm.clj:410)
	shadow.build.targets.esm/flush-dev (esm.clj:406)
	shadow.build.targets.esm/process (esm.clj:530)
	shadow.build.targets.esm/process (esm.clj:511)
	clojure.lang.Var.invoke (Var.java:384)
	shadow.build/process-stage/fn--17445 (build.clj:163)
	shadow.build/process-stage (build.clj:160)
	shadow.build/process-stage (build.clj:152)
	shadow.build/flush (build.clj:543)
	shadow.build/flush (build.clj:537)
	shadow.cljs.devtools.server.worker.impl/build-compile (impl.clj:369)
	shadow.cljs.devtools.server.worker.impl/build-compile (impl.clj:349)
	shadow.cljs.devtools.server.worker.impl/eval19004/fn--19006 (impl.clj:448)
	clojure.lang.MultiFn.invoke (MultiFn.java:234)
	shadow.cljs.devtools.server.util/server-thread/fn--18762/fn--18763/fn--18771 (util.clj:283)
	shadow.cljs.devtools.server.util/server-thread/fn--18762/fn--18763 (util.clj:282)
	shadow.cljs.devtools.server.util/server-thread/fn--18762 (util.clj:255)
	java.lang.Thread.run (Thread.java:833)
```

viewer/deps.edn
mhuebert: The build code we have depends directly on shadow-cljs though

sritchie:
I'm experimenting with this for local development and it seems great. This might fully kill my shadow-cljs config.

It looks like shadow-cljs has its file watcher enabled when this command runs: https://github.com/mhuebert/inside-out/blob/main/dev/user.clj#L11-L16

what files does this watch? If I add some entry to `:extra-namespaces` that depends on code in `src`, say, will shadow recompile when it sees changes there?

I THINK what happened before was that shadow watched `:src-paths`. here does it only watch extra-namespaces? It might be great to take filepaths or globs to watch so that I can do development against a local copy of a cljs library, for example.

mhuebert:
> what files does this watch? If I add some entry to `:extra-namespaces` that depends on code in `src`, say, will shadow recompile when it sees changes there?

AFAIK when used in embedded mode, shadow will use the classpath of your running JVM instance, so it respects whatever sources you've put in your own `deps.edn`, and shadow then watches files based on crawling the module-entries (where your `:extra-namespaces` are added). 

> so that I can do development against a local copy of a cljs library

for inside-out I was doing this with an extra alias that would add a local copy of this repo:
https://github.com/mhuebert/inside-out/blob/2f1cc36b343fa3b0f2531ce6799c3f1319b81c35/deps.edn#L21

mhuebert:
> Getting this failure locally after the merge on `bb dev`:

Fixed in f18afd7.

Breaking change: renamed `release!` to `release`.

@mk re: structure, I moved the viewer build into `nextjournal.clerk.builder.cljs` and put both release + watch there, & left the `dev-launcher` ns for compatibility.

viewer/src/nextjournal/clerk/viewer/builder.clj
```diff
@@ -0,0 +1,37 @@
+(ns nextjournal.clerk.viewer.builder
```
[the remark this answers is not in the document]
mhuebert: done

viewer/src/nextjournal/clerk/viewer/builder.clj
```diff
@@ -0,0 +1,37 @@
...
+               (shadow/release*
+                (get-build opts)
+                shadow-opts))
+        _ (when-not server-running? (shadow.server/stop!))
+        output-name (->> (or (:shadow.build.closure/modules state)
+                             (:build-modules state))
+                         (filter (comp #{:viewer} :module-id))
+                         first
+                         :output-name)]
+    (str "/js/" output-name)))
```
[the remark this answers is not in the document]
mhuebert: I've done this now - but the html file still references `_data` directly rather than a bucket somewhere.

viewer/src/deps.cljs
```diff
@@ -0,0 +1,27 @@
+{:npm-deps
+ {"@codemirror/autocomplete" "^6.0.2"
```
sritchie, line 2: Based on my local install with these deps, on main I think `@nextjournal/lang-clojure` is still missing

mk:
This has been supported by the `:cljs-namespaces` option in https://github.com/mentat-collective/clerk-utils, see https://clerk-utils.mentat.org/#static-build.

PR #262, opened 2022-11-03
Add `render` macro for rendering directly to Clerk notebook

sritchie:
(This macro was inspired by @mhuebert 's work on https://inside-out.matt.is/.)

This PR adds a macro that lets you write concise code in the case where you want to render some Reagent code directly in the notebook.

This comes up in cases where, for example, some other component is modifying a ratom and you want to render some view of it without any JVM input.

I'm not sure on the final name. The original macro from @mhuebert was called `cljs`.

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -334,7 +334,7 @@
     `(clerk/with-viewer v/examples-viewer
        (mapv (fn [form# val#] {:form form# :val val#}) ~(mapv (fn [x#] `'~x#) body) ~(vec body)))))
 
-(defmacro cljs
+(defmacro render
   "Returns a value that treats the supplied forms in the rendered Clerk notebook
   as if they were the body of a `:render-fn` with no arguments.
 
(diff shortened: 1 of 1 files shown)


zampino:
Nice idea! This more concise version 
```clojure
(defmacro cljs [& body]
  `(clerk/with-viewer '(fn [_] (do ~@body)) nil))
```
works against latest main branch, where you no longer need `v/html` calls if you return a vector.

sritchie:
Excellent. Cc @mhuebert who had a different version against the latest master, with a few tweaks to fix some issue with interactive components. I need to upgrade too to give this simpler version a try.

sritchie:
This was the final @mhuebert edition from inside-out:

```clj
(defmacro cljs [& exprs]
  `(clerk/with-viewer
     {:render-fn '(fn [_# _#]
                    (nextjournal.clerk.render/render-reagent
                     [(fn []
                        (let [result# (do ~@exprs)]
                          (if (vector? result#)
                            result#
                            [nextjournal.clerk.render/inspect result#])))]))}
     {}
     nil))
```

mk:
With #357 we have `eval-cljs` based on sci now. For regular cljs, I'd envision being able to switch between sci, regular cljs (and potentially others) but without adding to the api so I think this is best kept in a library for now.

PR #270, opened 2022-11-09
Make result viewer overridable again

zampino:
* leave result viewer entry-point open via regular viewer api
* fix card showing Clerk parser API client-side  
* fix static app


notebooks/cards.clj
```diff
@@ -199,13 +201,17 @@
...
+     doc
+     (nextjournal.clerk.parser/parse-clojure-string {:doc? true} doc)
+     (update doc :blocks (partial map (fn [{:as b :keys [type text]}]
+                                        (cond-> b
+                                          (= :code type)
+                                          (assoc :result
+                                                 {:nextjournal/value
+                                                  (let [val (eval (read-string text))]
+                                                    ;; FIXME: this won't be necessary once we unify v/html in SCI env to be the same as in nextjournal.clerk.viewer
+                                                    ;; v/html is currently html-render for supporting legacy render-fns
```
zampino, line 213: 👆 I'd prefer to fix this in a separate commit. It seems to me that legacy render-fn which use `v/html` in their body do make the browser hang if we unify the behaviour of the html viewer with n.clerk.viewer.

Issue #247, opened 2022-10-27
Blows up when using `potemkin/def-map-type`


pfeodrippe:
`potemkin/def-map-type` (https://github.com/clj-commons/potemkin#def-map-type) blows up for Clerk. `deftype` seems to work fine though. 

```clojure
The var `XXX` exists at runtime,
   but Clerk cannot find it in the namespace. Did you remove it?
```

Is there a way that I can say that Clerk shouldn't worry about this for a namespace or some extension point where, maybe, we could leverage clj-kondo information about this?

Using `declare` seems to workaround this.

FelipeCortez:
got the same error for hugsql's query vars

mk:
@pfeodrippe Thanks a lot for the report. When a Clojure file starts with an ns form, Clerk is checking if it has seems all dependency vars so it can warn folks when a var still exists at runtime but has been removed from the file.

This is going wrong here and #248 is a similar failure.

You should be able to opt out of this by having a first form that’s something other than a ns (e.g. an empty map) until we figure out if we can fix this or should provide an opt-out.

@FelipeCortez thanks for the pointer to the hugsql failure!

zampino:
Trying to have a look at this and it seems more complicated than expected. There's some stateful step that [potemkin is performing while macroexpanding](https://github.com/clj-commons/potemkin/blob/master/src/potemkin/types.clj#L302-L308) which messes up Clerk analysis. 

In a simple example like this

```clojure
(ns scratch.potemkin
  (:require [nextjournal.clerk :as clerk]
            [potemkin.types]
            [potemkin.collections :as c]))

(c/def-map-type M [m]
  (get [_ k dv] (clojure.core/get m k dv))
  (assoc [_ k v] (M. (assoc m k v)))
  (dissoc [_ k] (M. (dissoc m k)))
  (keys [_] (keys m)))

(->M {:a 1 :b 2})
```

the very first time Clerk shows this ns, analysis and eval goes fine. But not a second time 😕: potemkin macro `deftype+`  (see linked code) won't emit any deftype code and our analyzer won't be aware of the `->M` name, hence the warning: 

```
Execution error (ExceptionInfo) at nextjournal.clerk.analyzer/throw-if-dep-is-missing (analyzer.clj:252).
The var `#'scratch.potemkin/->M` exists at runtime, but Clerk cannot find it in the namespace. Did you remove it?
```

Now after a 
```clojure
(reset! potemkin.types/type-bodies {})
```
[...]

PR #272, opened 2022-11-09
Depth-first Analysis

zampino:
This is just an experiment around some possible changes in the order in which the dependency graph is built by Clerk analysis.

At present, when analyzing a single form, we collect only _immediate dependencies_ of the form. The full graph is computed only at a later time (immediately before evaluation) _and_ we don't follow/analyze dependencies which reside in jar deps.

The code in this branch changes the process in some ways:
* for each code block in the notebook, we do follow all dependencies transitively before passing to the next form
* we do parse and analyze parent documents (i.e. namespaces) residing in jars

with this in palce we can:

* eval (debatable) forms which transitively depend on `clojure.core/intern` 
* record the names of vars which have been interned during eval

to partly address #247, for instance.

There is an expected and relevant deterioration in performance here, since we do analyze (slurp/parse/analyze) foreign dependency namespaces regardless of them residing in jars. In order to mitigate this we can stop at certain forms belonging to known namespaces like `clojure.*` or `nextjournal.*` which are treaded as leaves in the tree and their jar should be hashed instead as we did before.

### DIFF
--- a/notebooks/interns.clj
+++ b/notebooks/interns.clj
@@ -1,0 +1,8 @@
+;; # 🎭 Manually interned Vars
+(ns interns)
+
+(defn my-intern [x] (intern *ns* x 42))
+
+(my-intern 'the-answer)
+
+the-answer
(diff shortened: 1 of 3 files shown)


zampino:
Reproducing analyzer issues in https://github.com/nextjournal/clerk-analyzer-issues.

zampino:
Superseded by #301.

PR #273, opened 2022-11-09
Refactor error boundary and result viewer

mk:
Drop the `!error` atom from the ErrorBoundary and provide a `use-error-handler` hook instead. Simplify the result viewer.


mk:
There's currently still an issue with a global error not disappearing without a reload, need to fix it before this can be merged.

PR #277, opened 2022-11-14
functional handling of resource-urls

mhuebert:
Serve and build commands accept a `:resource-urls` map which are merged into a `static-resource-urls` map (deprecating the `config/!resource->url` atom)

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -423,13 +423,14 @@
   Passing at least one of the above is required. When both `:paths`
   and `:paths-fn` are given, `:paths` takes precendence.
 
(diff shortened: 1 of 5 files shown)


mk (commented):
Looking over this I think I'd prefer a smaller non-breaking change that keeps the `!resource->url` atom as a global setting but allows to pass a functional override for the static build, maybe also on a doc level.

mhuebert:
Should be easy enough. We can stop using the atom entirely in this lib, but always read from the atom first, so it only comes into play if someone opts in to that behaviour.

zampino:
Trying to extract a more minimal version of this in #333.

PR #280, opened 2022-11-15
Accept :serve-paths option for local server

mhuebert:
Option to serve additional directories from Clerk's webserver.


mk:
I'd prefer Clerk's Webserver to only be for serving Clerk, as it's very basic. For doing custom stuff I'd prefer it if folks used their own server.

Note that with 70455ceab2ee53c3775980113bfa1ed8313a46cb we're now on a shadow-cljs version that sets the `Access-Control-Allow-Origin` `*` header so you should be able to use shadow-cljs for serving the custom js build.

Let me know if this wasn't the problem you were trying to solve here.

PR #285, opened 2022-11-17
Editable Code Viewer

zampino:
* Expose `nextjournal.clerk.render.code` namespace
* Add components to render editable code in render functions
* Expose codemirror view class by `codemirror.view` alias in SCI env


zampino:
This should be good now.

src/nextjournal/clerk/sci_env.cljs
```diff
@@ -115,7 +117,9 @@
 
 (def ^:export mount render/mount)
 
-(sci.ctx-store/reset-ctx! (sci/init initial-sci-opts))
+(sci.ctx-store/reset-ctx! (doto (sci/init initial-sci-opts)
+                            (sci/add-class! 'codemirror.view codemirror-view)))
```
mk, line 121: Could we also expose this under the original name `"@codemirror/view" in a follow-up? Maybe a question for @borkdude.

borkdude: AFAIK you can do `{:namespaces {"@codemirror/view ...}}` and then `(require '["@codemirror/view")` or do this via `:load-fn` and check for this string.

PR #291, opened 2022-11-21
Reset synced atom state in browser in connected clients

zampino:
Allow to propagate JVM state updates of synced atoms to clients via intentional/manual recompute.


ui_tests/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "nbb": "^1.0.138",
+    "nbb": "^1.1.147",
```
zampino, line 3: 👆 @borkdude fixed some issues which prevented yarn to correctly load nbb.

src/nextjournal/clerk/webserver.clj
```diff
@@ -103,17 +103,15 @@
...
                                     (create-ns 'user))]
-                   (let [{:as msg :keys [type]} (read-msg edn-string)]
+                   (let [{:as msg :keys [type sender-id]} (read-msg edn-string)]
                      (case type
                        :eval (do (eval (:form msg))
                                  (eval '(nextjournal.clerk/recompute!)))
                        :swap! (when-let [var (resolve (:var-name msg))]
                                 (apply swap! @var (eval (:args msg)))
-                                (doseq [ch @!clients :when (not= sender-ch ch)]
-                                  (httpkit/send! ch edn-string))
```
zampino, line 113: I believe this to be redundant now that we always reset synced atoms within `n.clerk.render/intern-atoms!`. This happens by broadcasting the whole document to all clients via the recompute below.

zampino:
Pushed [another approach](https://github.com/nextjournal/clerk/compare/downstream-sync...downstream-sync--drop-client-ids?expand=1) on top of this to avoid passing client-ids around but have the server decide based on sender channel equality.

PR #293, opened 2022-11-22
Fix var-from-def

zampino:
* Fix showing viewer maps with ::clerk/var-from-def keys
* Use unqualified key `:var-from-def?` in viewer maps to opt-out of extracting values from defined vars


mk:
This is not working:
Screenshot 2022-11-23 at 09 50 04

Issue #217, opened 2022-09-22
Changing ::clerk/width does not report spec failures for invalid values


dgtized:
Following through in the workshop, in https://github.com/nextjournal/clerk-demo/blob/master/workshop/exercises/ex_2_table_viewer.clj#L45:

changing `::clerk/width :full` to `::clerk/width :foo` does not report an error for invalid values .

In contrast, invalid values for `:code` in `::clerk/visibility {:code :fold}` report spec failures on invalid states.

Thanks!

mk:
Closed by #294.

PR #294, opened 2022-11-22
Viewer class customization

joe-loco:
This is a first draft for providing custom classes to viewers and the notebook viewer which should allow for most use cases and does not require actually overriding the base styles. Once a `:nextjournal.clerk/css-class` is available on the viewer or in document settings, the available class will be used and no further viewer classes will be assigned.

Here are a few simple class-based customizations:

https://user-images.githubusercontent.com/1944/203326797-9cbb5a19-5823-42f9-8cc6-c64a80e0d203.mp4

CleanShot 2022-11-22 at 16 12 49@2x


joe-loco:
closes #217

PR #299, opened 2022-11-28
Fix image src path when using custom index

Sohalt:
Fixes https://clojurians.slack.com/archives/C04ATSH4QNM/p1669658937784419

### DIFF
--- a/src/nextjournal/clerk/viewer.cljc
+++ b/src/nextjournal/clerk/viewer.cljc
@@ -383,15 +383,19 @@
        cas-url)))
 
 #?(:clj
-   (defn relative-root-prefix-from [index path]
-     (if (= index path) "" (str/join (repeat (get (frequencies (str path)) \/ 0) "../")))))
+   (defn relative-root-prefix-from [path]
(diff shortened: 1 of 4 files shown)


src/nextjournal/clerk/viewer.cljc
```diff
@@ -387,11 +387,15 @@
...
+   (defn map-index [{:as _opts :keys [index]} path]
+     (get {index "index.clj"} path path)))
+
+#?(:clj
+   (defn maybe-store-result-as-file [{:as doc+blob-opts :keys [index out-path file]} {:as result :nextjournal/keys [content-type value]}]
      ;; TODO: support customization via viewer api
      (if-let [image-type (second (re-matches #"image/(\w+)" content-type))]
        (assoc result :nextjournal/value
-              (str (relative-root-prefix-from file)
+              (str (relative-root-prefix-from (map-index doc+blob-opts file))
```
mk, line 398: Slightly prefer this to the second index arg to `relative-root-prefix-from` as we can later grow map-index into a general routing function.

PR #302, opened 2022-11-29
Improve error handling for `:paths-fn`

Sohalt:


src/nextjournal/clerk/builder.clj
```diff
@@ -132,16 +132,20 @@
...
-                                                   (fn? @resolved-var) (apply []))]
-                              (when-not (sequential? resolved-paths)
-                                (throw (ex-info (str "#'" paths-fn " must be sequential.") {:paths-fn paths-fn :resolved-paths resolved-paths})))
-                              resolved-paths)
-                            (throw (ex-info (str "#'" paths-fn " cannot be resolved.") {:paths-fn paths-fn}))))
-                        (throw (ex-info "`:path-fn` must be a qualified symbol pointing at an existing var." {:paths-fn paths-fn}))))
+             paths-fn (let [ex-msg "`:path-fn` must be a qualified symbol pointing at an existing var."]
+                        (when-not (qualified-symbol? paths-fn)
+                          (throw (ex-info ex-msg {:paths-fn paths-fn})))
+                        (let [resolved-var  (try (requiring-resolve paths-fn)
```
Sohalt, line 138: `requiring-resolve` returns nil if the namespace exists but the var doesn't.
The nullpointer will get caught by the catch in line 143, but the error message could be better in that case imho.
I think that should also result in `"`:path-fn` must be a qualified symbol pointing at an existing var."

test/nextjournal/clerk/builder_test.clj
```diff
@@ -37,10 +38,14 @@
...
-    (is (thrown? Exception (builder/expand-paths {})))
-    (is (thrown? Exception (builder/expand-paths {:paths-fn 'foo})))
-    (is (thrown? Exception (builder/expand-paths {:paths-fn "hi"})))
-    (is (thrown? Exception (builder/expand-paths {:index ["book.clj"]})))))
+    (is (thrown? ExceptionInfo (builder/expand-paths {})))
+    (is (thrown? ExceptionInfo (builder/expand-paths {:paths-fn :foo})))
+    (is (thrown? ExceptionInfo (builder/expand-paths {:paths-fn 'foo})))
+    (is (thrown? ExceptionInfo (builder/expand-paths {:paths-fn 'foo/bar})))
+    (is (thrown? ExceptionInfo (builder/expand-paths {:paths-fn 'clojure.core/inc})))
+    (is (thrown? ExceptionInfo (builder/expand-paths {:paths-fn 'clojure.core/+})))
```
Sohalt, line 46: Add a case for a non existing var in an existing namespace? e.g. `'clojure.core/this-var-does-not-exist`

mk:
Test failures unrelated (due to an outage of a server from wich we load images).

src/nextjournal/clerk/builder.clj
```diff
@@ -132,16 +132,21 @@
...
-                              resolved-paths)
-                            (throw (ex-info (str "#'" paths-fn " cannot be resolved.") {:paths-fn paths-fn}))))
-                        (throw (ex-info "`:path-fn` must be a qualified symbol pointing at an existing var." {:paths-fn paths-fn}))))
+             paths-fn (let [ex-msg "`:path-fn` must be a qualified symbol pointing at an existing var."]
+                        (when-not (qualified-symbol? paths-fn)
+                          (throw (ex-info ex-msg {:paths-fn paths-fn})))
+                        (if-let [resolved-var  (try (requiring-resolve paths-fn)
+                                                    (catch Exception e
+                                                      (throw (ex-info ex-msg {:paths-fn paths-fn}))))]
+                          (let [resolved-paths (try (cond-> @resolved-var
```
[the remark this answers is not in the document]
mk: That's because we also support it being a `def`, not only a defn.

PR #319, opened 2022-12-05
Preserve Whitespace in string Results

### LINKED ISSUE #314: Spaces are missing in output strings
If I have a string with any leading spaces, or multiple consecutive spaces, they are removed when showing the output. For example, this:
```clojure
;; A string with spaces:
"    hello     there     "
```
It is rendered as:
image
The leading spaces are removed, and all groups of more than one space are reduced to a single space.

I'm not sure if this is related to #306, but it feels different, so I opened a separate issue.

zampino:
Fixes #314.

### DIFF
--- a/src/nextjournal/clerk/render.cljs
+++ b/src/nextjournal/clerk/render.cljs
@@ -415,15 +415,13 @@
 
 (defn render-quoted-string [s {:as opts :keys [path viewer !expanded-at] :or {path []}}]
   (let [{:keys [closing-paren]} viewer]
-    [:span.cmt-string.inspected-value.inline-flex
-     [:span
-      (if (some #(and (string? %) (str/includes? % "\n")) (if (string? s) [s] s))
-        [expand-button !expanded-at "\"" path]
-        [:span "\""])]
-     [:div
-      (viewer/->value (render-string s opts))
-      "\""
-      closing-paren]]))
+    [:span.cmt-string.inspected-value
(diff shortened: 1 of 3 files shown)


src/nextjournal/clerk/render.cljs
```diff
@@ -414,16 +414,12 @@
           (if (string? s) [s] s))))
 
 (defn render-quoted-string [s {:as opts :keys [path viewer !expanded-at] :or {path []}}]
```
zampino, line 416: Took the opportunity to simplify this, assuming presented nodes with string values will always give just a string value. Is this correct?

emlyn:
I just tried this, and it looks much better, but there is still an extra newline inserted at the start of the string (if you copy the rendered result and paste it somewhere), although it is not displayed in the browser.

zampino:
@emlyn thanks for testing this stuff :-) 

> there is still an extra newline inserted at the start of the string (if you copy the rendered result and paste it somewhere), although it is not displayed in the browser.

I'm afraid that's due to some markup that we need in order to display un/collapsed strings correctly. Especially when they're part of nested collections. But we'll investigate. 

As a partial workaround: if your selection doesn't span the surrounding quotes then the string should be copied as is.

emlyn:
The newline thing is really nitpicking, this is great. Thanks for the fast fixes!

mk:
@emlyn I've merged the PR as-is, think we'll handle copying of results by some other means in the future.

PR #324, opened 2022-12-08
Hide Clerk-specific metadata from code blocks

zampino:
This removes the Clerk-specific metadata annotation like `^{:nextjournal.clerk/viewer ,,,}` from the code displayed in code cells in order to not distract from the essence. Metadata not coming from Clerk is left intact.


test/nextjournal/clerk/analyzer_test.clj
```diff
@@ -227,7 +227,6 @@
 (str my-uuid)
 my-uuid"
                    analyze-string
-                   ana/analyze-doc
```
zampino, line 230: analyze-string already calls analyze-doc. 

Now, maybe it's a bad idea to make analyze-doc not idempotent 🤔. It is so atm because I'm updating `:text` on blocks during analyze-doc to be stripped of clerk metas.

jackrusher:
E si gioca secondo le regole del gioco 😹

test/nextjournal/clerk/analyzer_test.clj
zampino: Definitely need to have a new key!

PR #357, opened 2022-12-19
SCI env improvements

zampino:
* allow to require codemirror classes by their own name
* expose namespaces from clojure-mode


zampino:
The following example consistently work, when building the render-fns step-by-step and using `clerk/show!` but fail when reloading the browser: 

CleanShot 2022-12-19 at 19 21 37@2x

zampino:
Also doesn't work in the static app https://snapshots.nextjournal.com/clerk/build/0c3d838215278ac6b57b13457da774ec5b49d3c2/index.html#/notebooks/viewers/code.clj

src/nextjournal/clerk/sci_env.clj
```diff
@@ -0,0 +1,8 @@
+(ns nextjournal.clerk.sci-env)
+
+(defmacro sci-copy-nss [& nss]
+  (into {} (for [[_ ns] nss]
+             [(list 'quote ns) `(sci.core/copy-ns ~ns (sci.core/create-ns '~ns))])))
```
mk, line 5: @borkdude any interest in getting such a convenience macro or something similar into sci?

borkdude:
@zampino The errors probably have to do with non-top-level require and refer. Perhaps it works if you execute the require and usage of that require in two separate functions, but ultimately this is what `eval-cljs-str` was intended for.

src/nextjournal/clerk/sci_env.clj
mk: or would it be possible to write this as a function if `sci.core/copy-ns` also accepted a quoted symbol as the first arg?

borkdude: I don't intend to add this macro today since it hides the ability to pass in the result of `sci/create-ns` which can carry metadata etc.

You have to do this using a macro since `sci.core/copy-ns` is also a macro.

mk: or if not a quoted symbol, maybe a namespace:
```clojure
(into {}
      (map (fn [ns-sym] [ns-sym (sci.core/copy-ns (find-ns ns-sym) (sci.core/create-ns ns-sym))]))
      '[nextjournal.clerk.parser nextjournal.clerk.render])
```

zampino:
> non-top-level require

I see. But also using requires in a separate render-fns doesn't survive a page reload: 

CleanShot 2022-12-20 at 08 49 08@2x

zampino:
>  this is what `eval-cljs-str` was intended for.

yeah, that works

CleanShot 2022-12-20 at 08 56 24@2x

PR #368, opened 2022-12-22
Fix js-interop destructuring in SCI context

borkdude:
Screenshot 2022-12-22 at 11 59 56


zampino (approved):
🙏 @borkdude! I'll merge after bumping sci configs.

PR #370, opened 2022-12-23
Fix hashing on windows

mk:
This fixes the `nextjournal.clerk.render.hashing/front-end-hash` computation that would previously fail on windows. Also set `.gitattributes` to get the same line endings on windows so the hash is stable on different platforms.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+;; always force unix line endings to make hashing consistent
+* text eol=lf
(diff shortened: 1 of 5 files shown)


respatialized:
Running the most recent commit in this PR locally results in a `FileNotFoundException` for the GCS lookups to `storage.googleapis.com` for me.

mk:
@respatialized if you can, I'd appreciate if you could try this once more, using the latest sha on this PR as a git dependency and confirm that

```clojure
(= "3mXdcLQSNAEFTLdDXoAw47rKytXS"
   ((requiring-resolve 'nextjournal.clerk.render.hashing/front-end-hash)))
;;=> true
```

respatialized:
Yes, I am able to load `clerk` and that expression evaluates to `true` on Windows. ✔️

PR #373, opened 2022-12-25
Add emoji-regex

sritchie:
Without this I get

```
found 0 vulnerabilities
[:clerk] Compiling ...
The required JS dependency "emoji-regex" is not available, it was required by "nextjournal/clerk/render/navbar.cljs".

Dependency Trace:
	nextjournal/clerk/static_app.cljs
	nextjournal/clerk/render.cljs
	nextjournal/clerk/render/navbar.cljs

Searched for npm packages in:
	/Users/sritchie/code/clj/mathbox.cljs/node_modules
```

### DIFF
--- a/render/deps.cljs
+++ b/render/deps.cljs
@@ -14,6 +14,7 @@
             "@nextjournal/lang-clojure" "1.0.0"
             "@nextjournal/lezer-clojure" "1.0.0"
             "d3-require" "^1.2.4"
+            "emoji-regex" "10.2.1"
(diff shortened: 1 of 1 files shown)


zampino:
Maybe we could drop the emoji-extraction functionality from the navbar library [now that markdown can do it](https://github.com/nextjournal/markdown/pull/9).

mk:
@sritchie we've gone with the alternative of dropping the unused require. Can you confirm this fixes your issue as well?

PR #379, opened 2023-01-02
Drop emoji-regex require

zampino:
Closes #373.

### DIFF
--- a/src/nextjournal/clerk/render/navbar.cljs
+++ b/src/nextjournal/clerk/render/navbar.cljs
@@ -1,12 +1,9 @@
 (ns nextjournal.clerk.render.navbar
-  (:require ["emoji-regex" :as emoji-regex]
-            ["framer-motion" :as framer-motion :refer [motion AnimatePresence]]
+  (:require ["framer-motion" :as framer-motion :refer [motion AnimatePresence]]
             [nextjournal.clerk.render.localstorage :as localstorage]
             [applied-science.js-interop :as j]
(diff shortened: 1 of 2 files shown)


src/nextjournal/clerk/render/navbar.cljs
```diff
@@ -74,10 +71,8 @@
     (into
      [:div]
      (map-indexed
-      (fn [i {:keys [path title expanded? loading? items toc]}]
-        (let [label (or title (str/capitalize (last (str/split path #"/"))))
-              emoji (when (zero? (.search label emoji-re))
-                      (first (.match label emoji-re)))]
+      (fn [i {:keys [emoji path title expanded? loading? items toc]}]
+        (let [label (or title (str/capitalize (last (str/split path #"/"))))]
```
zampino, line 75: It seems to me that we never call `navbar-items` at runtime in the current usage of navbar in Clerk. @joe-loco could you have a look? We're passing a `:toc` to the navbar state and not `:items`.

joe-loco: I discussed this with @mk in Berlin. I initially wanted to drop all code that was not relevant to `:toc` but @mk wanted to keep it in for when Clerk is able to show hierarchies that are above notebooks.

joe-loco: Can discuss further when I’m back tmw if you want.

PR #381, opened 2023-01-06
Make `build-graph` recur until all transitive deps are analyzed

### LINKED ISSUE #375: Latest clerk errors on notebook analyses, ->hash must be ifn?
On https://github.com/mentat-collective/mathbox.cljs, branch `sritchie/new_release`, when I build my static bundle with `bb publish-local` I see the following output:

```
🔬 Analyzing… ->hash must be `ifn?` {:->hash nil, :codeblock {:form (show-sci [mathbox/Mathbox {:style {:height 500px, :width 100%}, :init {:background-color 16777215, :camera-position [3 2 5], :scale 500}} [mb/Cartesian {} [mb/Transform4 {:matrix (let [a (/ (* (:projectionAngle (clojure.core/deref !state)) 2 Math/PI) 360.0)] [1 0 0 (Math/cos a) 0 1 0 0 0 0 1 (Math/sin a) 0 0 0 1])} [mb/Area {:rangeX [-3 1], :rangeY [(* -2 Math/PI) (* 2 Math/PI)], :width 129, :height 65, :channels 4, :expr (fn [emit x y _i _j] (let [r (Math/exp x)] (emit (* r (Math/cos y)) (* r (Math/sin y)) x y)))}] [mb/Surface {:color (:surface (clojure.core/deref !state)), :zBias -0.25}] [mb/Line {:color (:line (clojure.core/deref !state)), :width 4}] [mb/Transpose {:order yxzw}] [mb/Line {:color (:line (clojure.core/deref !state)), :width 4}]]]]), :ns-effect? false, :freezable? true, :no-cache? false, :deps #{nextjournal.clerk/with-viewer mentat.clerk-utils.show/show-sci nextjournal.clerk/mark-presented}, :file /Users/sritchie/.gitlibs/libs/io.github.mentat-collective/clerk-utils/e40cdf79edf45748f5a2c6752c8156b3d240b087/src/mentat/clerk_utils/show.cljc, :id mathbox.examples.math.exponential/anon-expr-5drxcaGZr7zvK46gPq5a5FKTUYBvdG}}
Done in 600.672ms. ✅
```

This also comes up at the REPL when I save the file containing this form, but only sometimes.

I'm not sure how `:->hash nil` ends up in there. This is happening on version 5cf3982b29c25a6fe70ee4d4444293c5320ff697, and was NOT happening a week ago with the same code, though I haven't done a proper bisection to figure out when this came in.

philomates:
Until now Clerk did not analyze the full transitive dependency graph which could lead to Clerk not detecting a change properly. Analysis is now recursive which means it's taking a bit longer initially. We cache analysis results per file in memory so subsequent analysis should be fast. We will follow up with visualizing the progress of analysis & execution.

Also discovered cases where classes instead of symbols could end up in the dependency graph and introduced normalization to symbols.

This also gets rid of the `->hash must be ifn?` warning which fixes #375.

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -380,8 +380,9 @@
 
 #_(unhashed-deps {#'elements/fix-case {:deps #{#'rewrite-clj.node/tag}}})
 
-(defn ns->path [ns]
-  (str/replace (namespace-munge ns) "." fs/file-separator))
+(defn ns->path
+  ([ns] (ns->path fs/file-separator ns))
+  ([separator ns] (str/replace (namespace-munge ns) "." separator)))
 
 (defn ns->file [ns]
   (some (fn [dir]
@@ -393,7 +394,7 @@
         (cp/classpath-directories)))
 
 (defn ns->jar [ns]
-  (let [path (ns->path ns)]
+  (let [path (ns->path "/" ns)]
     (some #(when (or (.getJarEntry % (str path ".clj"))
                      (.getJarEntry % (str path ".cljc")))
              (.getName %))
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -20,6 +20,8 @@
             [taoensso.nippy :as nippy]
             [weavejester.dependency :as dep]))
 
+(set! *warn-on-reflection* true)
+
 (defn deref? [form]
   (and (seq? form)
        (= (first form) `deref)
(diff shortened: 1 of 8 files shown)


src/nextjournal/clerk/analyzer.clj
```diff
@@ -419,15 +461,54 @@
...
+
+         (-> (cond-> doc
+               (not (:graph doc)) analyze-doc)
+             (assoc :analyzed-file-set (cond-> #{} (:file doc) (conj (:file doc))))
+             (assoc :counter 0))]
+    (let [unhashed (unhashed-deps ->analysis-info)
+          loc->syms (apply dissoc
+                           (group-by find-location unhashed)
+                           analyzed-file-set)]
+      (prn :build-graph counter :analyzed-file-set analyzed-file-set)
```
borkdude, line 474: There is still a prn here, is that intended?

src/nextjournal/clerk/analyzer.clj
```diff
@@ -419,15 +461,54 @@
...
+      (if (and (seq loc->syms) (< counter 10))
+        (recur (-> (reduce (fn [g [source symbols]]
+                             (prn :source source :symbols symbols :ana (when (or (nil? source)
```
borkdude, line 477: Also one here

src/nextjournal/clerk/analyzer.clj
mk: added those to debug windows on CI. Will remove them before merging.

borkdude:
@mk I pushed a fix for Windows. I also added an alias for the cognitect test runner which I find easier to work with, I hope that's ok.

PR #392, opened 2023-01-23
Sidenotes

zampino:
This aims at getting a more solid and flexible handling of foot/sidenotes.

- [x] Fix sidenotes in paragraphs with marks
- [x] Sidenotes on narrow viewports
- [x] Keep sidenote ref count across blocks. 
- [x] Awaits markdown release after https://github.com/nextjournal/markdown/pull/11

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -10,7 +10,7 @@
 
         io.github.nextjournal/markdown {#_#_
                                         :local/root "../markdown"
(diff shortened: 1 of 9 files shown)


zampino:
CleanShot 2023-01-24 at 09 54 56@2x

This is currently happening with Chrome Version 101.0.4951.64 (which is not the latest).

PR #393, opened 2023-01-23
VS Code clerk/show!

brettrowberry:
Add VS Code clerk/show! file watcher hotkey instructions.

### DIFF
--- a/book.clj
+++ b/book.clj
@@ -130,7 +130,25 @@
 ;; nmap <silent> <localleader>cs :execute ClerkShow()<CR>
 ;; ```
 
+;; **VS Code/Calva**
 
+;; In VS Code/Calva, see [Custom REPL Commands](https://calva.io/custom-commands/).
+;; Here are the snippets you'll need (for both serving and showing):
+
+;; ```json
+;;     "calva.customREPLCommandSnippets": [
+;;        {
+;;            "name": "Clerk Serve",
+;;            "key": "v",
+;;            "snippet": "(nextjournal.clerk/serve! {:browse? true})"
+;;        },
+;;        {
+;;            "name": "Clerk Show",
+;;            "key": "h",
+;;            "snippet": "(nextjournal.clerk/show! \"$file\")"
+;;        }
+;;    ]
+;; ```
 
 ;; ## 🔍 Viewers
 


brettrowberry:
Thanks @mike706574 for helping me figure this out!

brettrowberry:
Hmm, looks like this info is also in https://github.com/nextjournal/clerk#editor-workflow so I'd need to update that too.

mk:
Hey @brettrowberry,

Thanks a for looking into this, been wanting to have this for a while. Do you think setting up a global shortcut for `nextjournal.clerk/show!` would also be reasonable? I use `alt+enter` in Emacs. I'm on the fence if we should include a shortcut for `nextjournal.clerk/serve!`, maybe leave that as an exercise for the reader?

blnote:
the show! shortcut of book.clj does not work for me in windows, because $file contains escaped characters "C:\U...." that can not be read by the REPL
easiest fix would probably be on Calva's side by escaping the path

PEZ:
> $file contains escaped characters "C:\U...." that can not be read by the REPL

Ouch! That needs to be fixed in Calva. Please file an issue there!

mk:
Thanks for looking into this @PEZ, I've filed https://github.com/BetterThanTomorrow/calva/issues/2184.

PEZ:
Thanks for the issue, @mk!

@blnote : Here's a build of Calva that should be escaping the file path backslashes. I don't have a Windows machine handy, can you give it a spin?

* https://output.circle-artifacts.com/output/job/3f508067-675f-493b-8d6b-f75735ce2464/artifacts/0/tmp/artifacts/calva-2.0.358-2184-escape-windows-filenames-d34b218a.vsix

To install it you need to download the VSIX package to your machine (and rename it the extension to `.vsix` if it downloads as a `.zip`), then from the Extension pane in VS Code, select **Install from VSIX...** in the meatballs menu at the top of the pane.

PEZ:
This fix is now released, blindly, in Calva v2.0.358. All that is changed is that all `\` in `$file` will be replaced with `\\`. So, it really _should_ fix the issue...

Another thing. I'd like to mention a new feature in Calva, released with v2.0.356, that might, or might not, be relevant for Clerk notebooks workflow. It's a way to get code called when the file is loaded in the REPL (using Calva's command for **Load File**). The same substitutions as with the custom command snippets are available here. See: [Calva: Auto-evaluate Code at file/namespace load/evaluation](https://calva.io/customizing-jack-in-and-connect/#auto-evaluate-code-at-filenamespace-loadevaluation) for details.

blnote:
Hi Peter, I've sent u an email yesterday but didn't reply to the list,
maybe it got lost. I tried it and the issue was fixed for me, so the
simple replacement worked. Thanks again, Bastian

Am 09.05.2023 um 11:25 schrieb Peter Strömberg:

brettrowberry:
I haven't used Clerk for a long time. Feel free to close of do whatever. Thanks!

PR #397, opened 2023-02-01
Update README.md

davesann:
added notes regarding watch-paths and neovim

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,13 @@ endfunction
 
 nmap <silent> <localleader>cs :execute ClerkShow()<CR>
 ```
+##### Issues with neovim and watch-paths
+
+If using neovim and watch-paths, ensure that you:
+```
+:set backupcopy=yes
+```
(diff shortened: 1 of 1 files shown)


davesann:
This change to readme will no longer be needed (generally) if bumping version of https://github.com/gmethvin/directory-watcher to 0.17.3 or above in your beholder dependency. (see https://github.com/gmethvin/directory-watcher/issues/88).

Users of clerk that have reloading issues can bump this themselves by adding an explicit dependency.

mk:
Hi @davesann, thanks for looking into this. I've cut a new beholder release with the directory watcher bump and upgraded Clerk. Assume that fixes your issue, please let me know if there's anything else we need to do here.

davesann:
I included  ```com.nextjournal/beholder {:mvn/version "1.0.2"}```  and the problem is resolved for me using neovim. Note that there are still some editors for which there may be issues.

PR #400, opened 2023-02-02
Add register-js! for registering js libraries

sritchie:
Docstring as example:

Given some string `libname` and a JS namespace object (and an optional symbolic `alias`), makes the JS namespace available to SCI.

 For example, given a namespace with `(:require [\"react\" :as re])`:

```clj
(register-js! \"react\" re)
```

Makes it possible to call

```clj
(clerk/eval-cljs '(require '[\"react\" :as react]))
```

in a notebook.

Calling

```clj
(register-js! \"react\" re 'react)
```

makes the `react` alias available to every consumer of the SCI environment.

### DIFF
--- a/src/nextjournal/clerk/sci_env.cljs
+++ b/src/nextjournal/clerk/sci_env.cljs
@@ -155,6 +155,41 @@
 
 (def ^:export mount render/mount)
 
+(defn register-js!
+  "Given some string `libname` and a JS namespace object (and an optional symbolic
+  `alias`), makes the JS namespace available to SCI.
+
+  For example, given a namespace with `(:require [\"react\" :as re])`:
+
+  ```clj
+  (register-js! \"react\" re)
+  ```
+
+  Makes it possible to call
+
+  ```clj
+  (clerk/eval-cljs '(require '[\"react\" :as react]))
+  ```
+
+  in a notebook.
+
+  Calling
+
+  ```clj
+  (register-js! \"react\" re 'react)
+  ```
+
+  makes the `react` alias available to every consumer of the SCI environment.
+  "
+  ([libname js-namespace]
+   (set! libname->class
+         (assoc libname->class libname js-namespace)))
+  ([libname js-namespace alias]
+   (register-js! libname js-namespace)
+   (when alias
+     (eval-form
+      `(~'require '[~libname :as ~alias])))))
+
 (sci.ctx-store/reset-ctx! (sci/init initial-sci-opts))
 
 (sci/alter-var-root sci/print-fn (constantly *print-fn*))


mk:
Good catch that we didn't expose react. We already had a different way to expose existing javascript classes to sci so I've used that in the attached commit. You can also require react now but we don't need users to make the register call before hand. Let me know if this solves your original problem!

mk:
Maybe I shouldn't have picked this up before my first coffee. Looking at your commit again, you're going through `libname->class` so your problem is more about write access to this mapping to expose libs other than react? In that case I think I'm in favor of turning `libname->class` into an atom (plus maybe find a better name).

borkdude:
The goal of this PR was to let people add JS libraries to the SCI config. It was not about react specifically. Instead of the `eval-form` the helper function can also call `load-fn` directly.

mk:
@borkdude should turning `libname->class` into an atom as suggested above also enable letting people add JS libraries to SCI? There no sci concept for this we could use here, correct?

borkdude:
The SCI concept is already used in the load-fn.

borkdude:
If we gloss over what is currently in the `:load-fn`:

```
(defn load-fn [{:keys [libname ctx opts ns]}]
  (when (contains? libname->class libname)
    (let [{:keys [as refer]} opts
          munged-libname (symbol (munge libname))
          klass (libname->class libname)]
      (sci/add-class! ctx munged-libname klass)
      (when as
        (sci/add-import! ctx ns munged-libname as))
      (doseq [r refer]
        (when-some [prop (j/get klass (name r))]
          (let [sub-libname (str munged-libname "$" r)]
            (sci/add-class! ctx sub-libname prop)
            (sci/add-import! ctx ns sub-libname r))))
      {:handled true})))
```

To add a class by name into the SCI context you can use:

```
(sci/add-class! ctx munged-libname klass)
```

To add an import (here used for alias) to a namespace in the context you can use:

```
(sci/add-import! ctx ns sub-libname r)
```

mk:
I was more asking if we can unify how folks extend the sci env for regular namespaces vs js classes. For regular js namespaces you go through the global sci context. It might be nice to also go through this route for regular js classes to have this state in one place?

One idea might be to store the (mutable) libname->class map in the sci context?

borkdude:
I'm afraid I don't understand what you mean. Unify where, in clerk? What is "regular js namespaces" vs "regular js classes"? What do you mean with "go through the global sci context". I don't see why you need libname->class at all if you want to extend it yourself, but the above post by Sam was just a quick hack based on what was already present. With `sci/add-class!` and `sci/add-import` you can do it yourself as well.

mk:
I mean (sci) namespaces like `reagent.core` and js classes like `"@codemirror/api"` (how I would require them.

We currently provide sci namespaces when setting up the context, and js classes in load-fn. There's no place (afaik) currently to store a mapping of (unmunged) string names to js classes in the sci context. 

Or should I just add the munged symbol to `:classes` and things will work without the `libname->class` map in Clerk?

Not sure if this clarifies things. Also happy to talk about this on a call next week.

borkdude:
JS libs (called classes in SCI for unifying the concept between JVM and JS) and clojure namespaces are not unified since they are two different concepts. SCI leans more towards the JVM side here where vars are mutable objects with metadata, etc. things whereas classes are host objects that are not aware of Clojure. SCI has an API to add classes to the context and to define imports for a namespace. This is unified between JVM and JS but I don't have plants to unify "classes" with "namespaces" as classes have different semantics.

mk:
> clojure namespaces are not unified since they are two different concepts.

That's clear. I meant supporting passing string keys in `:classes` when initializing the sci context. See 7aac5bda214fa9b1481fb8bd12229bdd36493e00 in which I dropped the `libname->class` map and let `load-fn` use the already existent class. I think it might be nice if

1. sci supported string keys in `:classes` doing the `(symbol (munge libname))` conversion in that case
2. sci would support requiring string classes out of the box so we could pass these classes without needing a `:load-fn` in that case

Can do a PR if you think this is reasonable.

borkdude:
Right, experimental PR welcome!

PR #401, opened 2023-02-03
Set sci ns to mirror JVM ns

### LINKED ISSUE #362: Evaluate sci forms in same namespace as notebook?
I've noticed that `*ns*` is bound to `user` for all notebooks; can we set it to the namespace of the current notebook? This would make `^{::clerk/sync true}` feel very nice, since we wouldn't need to add a namespace prefix.

It would also prevent a case where evaluating a `defn` with the same name in different notebooks (imagine `->string` or something generic) will clash.

mk:
Closes #362.

### DIFF
--- a/notebooks/controlling_width.clj
+++ b/notebooks/controlling_width.clj
@@ -12,5 +12,5 @@
 (clerk/html {::clerk/width :full} [:div.w-100.bg-green-200.pl-2 ":full"])
 
 ;; This works with metadata as well.
-^{::clerk/width :wide ::clerk/viewer :html}
+^{::clerk/width :wide ::clerk/viewer 'nextjournal.clerk.viewer/html-viewer}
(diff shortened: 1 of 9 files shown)


mk:
UI tests are green now but there's still issues e.g. with the [slideshow viewer in the static build](https://snapshots.nextjournal.com/clerk/build/aa7db7678f646f5193938dca9e22dd5891675da9/index.html#/notebooks/slideshow.clj) but that doesn't trigger a ui test failure because the render error doesn't log an error to the console atom.

As we can see in the commits to fix the ui tests, this is currently a breaking change. I'm thinking about keeping it this way though, because I think on the flip side, things are much clearer in the long term because there's no more question about what the `v` alias points to.

mk:
Need to follow up with good docs about this.

PR #402, opened 2023-02-04
Bump sci with support for adding js libs using `sci/add-js-lib!`

mk:
This adds support for adding js libs to the sci context using newly added `sci/add-js-lib!` that can then be required. We can thus drop the `libname->class` map and the sci `:load-fn`.

This is an alternative to #400.


mk:
@sritchie do you have a use case you can test this with?

sritchie:
@mk @borkdude thank you, this is great!!

PR #407, opened 2023-02-06
Fix `example` macro

pfeodrippe:
Before, the user had to have `nextjournal.clerk` aliased to `clerk` in their namespace or they would have an exception.

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -346,7 +346,7 @@
   Does nothing outside of Clerk, like `clojure.core/comment`."
   [& body]
   (when nextjournal.clerk.config/*in-clerk*
-    `(clerk/with-viewer v/examples-viewer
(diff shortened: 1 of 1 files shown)


src/nextjournal/clerk.clj
```diff
@@ -346,7 +346,7 @@
   Does nothing outside of Clerk, like `clojure.core/comment`."
   [& body]
   (when nextjournal.clerk.config/*in-clerk*
-    `(clerk/with-viewer v/examples-viewer
+    `(nextjournal.clerk/with-viewer v/examples-viewer
```
mk, line 349:
```clojure
`(with-viewer ,,,
```
Would also work, right?

mk (approved):
Good catch, thanks @pfeodrippe!

src/nextjournal/clerk.clj
pfeodrippe: Yep :D thanks

PR #416, opened 2023-02-13
Drop parsing of hashtags and internal links in markdown

### LINKED ISSUE #383: Hyperlinks with a `#` are incorrectly parsed
I tried to insert the following link: [#2](https://github.com/mentat-collective/leva.cljs/issues/2)

and noticed that clerk resolves it as `2` without the `#`, and inserts a link to `#2` on the page:

image

If I add a SECOND pound sign, I get proper rendering, but the link still leads to `#2` on the page:

image

This doesn't happen on GitHub, so it seems like it's not a markdown spec problem.

zampino:
Closes #383.

Since we never agreed on what either #tags or [[internal links]] should point to when rendered as html, we decided to stop parsing such expressions by default and rely on context-specific customization instead. I will restore suitable viewers once it's clear how tags and wiki-style links should behave.

See https://github.com/nextjournal/markdown/pull/14 for more context and examples.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -192,14 +192,7 @@
     "@lezer/highlight" "^1.1.3"
     "@lezer/lr" "^1.3.0"
 
-"@lezer/lr@^1.0.0":
-  version "1.2.0"
-  resolved "https://registry.yarnpkg.com/@lezer/lr/-/lr-1.2.0.tgz#59aecafdbc15be63f918cf777f470dd17562f051"
-  integrity sha512-TgEpfm9br2SX8JwtwKT8HsQZKuFkLRg6g+IRxObk9nVKQLKnkP3oMh+QGcTBL9GQsfQ2ADtKPbj2iGSMf3ytiA==
-  dependencies:
-    "@lezer/common" "^1.0.0"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -192,14 +192,7 @@
...
 
-"@lezer/lr@^1.0.0":
-  version "1.2.0"
-  resolved "https://registry.yarnpkg.com/@lezer/lr/-/lr-1.2.0.tgz#59aecafdbc15be63f918cf777f470dd17562f051"
-  integrity sha512-TgEpfm9br2SX8JwtwKT8HsQZKuFkLRg6g+IRxObk9nVKQLKnkP3oMh+QGcTBL9GQsfQ2ADtKPbj2iGSMf3ytiA==
-  dependencies:
-    "@lezer/common" "^1.0.0"
-
-"@lezer/lr@^1.3.0":
+"@lezer/lr@^1.0.0", "@lezer/lr@^1.3.0":
```
zampino, line 195: Didn't seem right to have 2 separate top-level entries for @lezer/lr for actually compatible versions.

PR #419, opened 2023-02-14
Suppport local Files in Markdown Images

### LINKED ISSUE #223: Consider internalizing local markdown images
I noticed an issue when using local images to build a clerk module. Markdown-defined images only work if they are full URLs but not if they are local files. The way this is handled in R-markdown, I believe, is to encode the images as bas64 strings.

```clj
;;  this will render just fine ...
;; ![full-url-image](http://my.site.com/image.jpg)
;;
;;  ... but the URL for this image will be wrong and the image will not show in Clerk
;; ![local-Image](/resources/local.jpg)
```

zampino:
With these changes, image sources in markdown comments are transformed according to:

* `serve!`: the file path is served as-is by clerk webserver
* `(build! {:bundle true})` images are inlined as base64 encoded data
* `(build! {:bundle false})`: images are content addressed and stored in the `_data` folder

In addition, we now allow `clerk/image` to be called against local paths.

Closes #223.


src/nextjournal/clerk/viewer.cljc
```diff
@@ -516,7 +517,9 @@
 
 (defn with-block-viewer [doc {:as cell :keys [type]}]
   (case type
-    :markdown [(with-viewer `markdown-viewer (process-sidenotes doc (:doc cell)))]
+    :markdown [(with-viewer `markdown-viewer
+                 {:nextjournal.clerk/opts (select-keys doc [:index :file :bundle? :out-path])}
```
zampino, line 521: 👆 Not super happy with the selec-keys here, but in this case it's harder to pass down doc options along the markdown tree. 

Options will land in the presentation:

```clj
(-> @nextjournal.clerk.webserver/!doc
      viewer/notebook
      viewer/present
      viewer/->value
      :blocks)
=>
[{:nextjournal/value [:div.markdown-viewer
                      [:p
                       [:<> "what a piece "]
                       [#viewer-evalnextjournal.clerk.render/inspect-presented
                        {:nextjournal/value [:img.inline
                                             {:src "_blob//Users/amantini/dev/clerk/images/vera.jpg", :alt ""}],
                         :nextjournal/viewer {:name nextjournal.clerk.viewer/html-viewer,
                                              :render-fn #viewer-fnidentity,
                                              :hash "5du5eveS8XMqMDMYH5rpQ1bqryEJdg"}}]
                       [:<> " ahoi"]]],
  :nextjournal/opts {:file "/Users/amantini/dev/clerk/notebooks/scratch/debug.clj"},
  :nextjournal/viewer {:name nextjournal.clerk.viewer/html-viewer-,
                       :render-fn #viewer-fnidentity,
                       :hash "5du1K91pV2nRmpCG3vM8fBcDsgHWKs"}}]
```
unless we explicitely dissoc them in `process-blocks`, but I'd find that even worse.

PR #431, opened 2023-03-09
Fragments

### LINKED ISSUE #430: Support :nextjournal.clerk/width on viewers rendered within Tap Inspector
### Support :nextjournal.clerk/width on viewers rendered within Tap Inspector

**Background**

Ash: 

> I’ve some wide data that I’d like to tap> to the Tap Inspector
> but setting {::clerk/width :full} on the data’s viewers ( clerk/table , v/vl, etc.)
> seems to be ignored.
> Am I doing  it wrong, is it a bug, or does Tap Inspector not support :nextjournal.clerk/width? :thinking_face:

Martin: 

> doesn’t support it, yet. Issue welcome

zampino:
Allow for splicing a seq of values into the document as if produced by results of individual cells. Useful when programmatically generating content. One of the advantages of using this as opposed to `clerk/col` is that the former respects widths set on wrapped values.

CleanShot 2023-03-09 at 12 34 08@2x

## Todos: 
- [x] fix pagination inside fragments
- [x] make tap inspector use fragments (Closes #430)

### DIFF
--- a/src/nextjournal/clerk/viewer.cljc
+++ b/src/nextjournal/clerk/viewer.cljc
@@ -568,12 +568,16 @@
                                                                               (select-keys cell [:loc]))}
                       ;; TODO: display analysis could be merged into cell earlier
                       (-> cell (merge display-opts) (dissoc :result))))
+
               (or result? eval?)
-              (conj (with-viewer (if result?
-                                   (:name result-viewer)
-                                   (assoc result-viewer :render-fn '(fn [_] [:<>])))
-                      {:nextjournal/opts {:id (processed-block-id (str id "-result"))}}
-                      (assoc cell ::doc doc)))))))
+              (into (map-indexed (fn [i x]
+                                   (with-viewer (if result? (:name result-viewer) (assoc result-viewer :render-fn '(fn [_] [:<>])))
+                                     {:nextjournal/opts {:id (processed-block-id (str id "-" i "-result"))}}
+                                     (-> cell (assoc ::doc doc) (assoc :result x)))))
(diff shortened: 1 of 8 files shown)


src/nextjournal/clerk/viewer.cljc
```diff
@@ -575,8 +575,8 @@
                                      {:nextjournal/opts {:id (processed-block-id (str id "-" i "-result"))}}
                                      (-> cell (assoc ::doc doc) (assoc :result x)))))
                     (or (when-some [fgm (-> cell :result :nextjournal/value (get-safe :nextjournal.clerk/fragment))]
-                          (map #(hash-map :nextjournal/value %
-                                          :nextjournal/blob-id (-> cell :result :nextjournal/blob-id)) fgm))
+                          (map-indexed #(hash-map :nextjournal/value %2
+                                                  :nextjournal/blob-id (str (-> cell :result :nextjournal/blob-id) %1)) fgm))
```
zampino, line 579: Not sure faking blob ids is the right move here, but couldn't find a better trick. Tried to provide an unambiguous initial path for the presentation here but didn't lead to any good.

zampino:
Added the fragment based tap inspector to a new notebook, just to compare how it looked like. I am not sure the horizontal ruler above the value makes sense in presence of different widths: 

CleanShot 2023-03-09 at 18 00 38@2x

thoughts? @joe-loco

zampino:
Superseded by 03fac495c1d17e3c44a1862587d87e6980c5ccb2.

PR #439, opened 2023-03-18
Document that viewers can be embedded inside of hiccup code

dackerman:
I was struggling to layout some elements using clerk, and the only way I could see to compose elements from the Book of Clerk was by using `clerk/row` and `clerk/col`. However, I wanted more control over the style. From slack, I was told that you can embed viewers directly into hiccup code!

I didn't see it documented in the Book of Clerk, so I decided to just add a couple small mentions in the hiccup and "composing viewers" sections:

in the hiccup section:
screenshot-2023-03-18T07:56:29-07:00

in the composing viewers section:
screenshot-2023-03-18T07:56:39-07:00

### DIFF
--- a/book.clj
+++ b/book.clj
@@ -168,10 +168,16 @@
 ;; You can style elements, using [Tailwind CSS](https://tailwindcss.com/docs/utility-first).
 (clerk/html [:button.bg-sky-500.hover:bg-sky-700.text-white.rounded-xl.px-2.py-1 "✨ Tailwind CSS"])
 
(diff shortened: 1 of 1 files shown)


joe-loco:
@dackerman Thanks for the PR! I added some new examples in the hope of showing each feature’s respective qualities better. Also wanted something "more real" than the donut charts.

CleanShot 2023-03-23 at 16 29 43@2x
CleanShot 2023-03-23 at 16 29 53@2x

PR #442, opened 2023-03-22
Make clerk/doc-url links work in show mode 

zampino:


zampino:
Compared to #437 we're evaluating the notebook twice (one show and one recompute) due to the `clerk-eval` from the client. So we'd need to pass some option to skip the recompute.

It also has the disadvantage that we cannot show a notebook by entering the path `/#show=<path-to-notebook.clj>` in the browser URL and pressing Enter.

One advantage this _could_ have over #437 though, is making use of history pushState when going to different notebooks via `clerk/show!` but I couldn't have it reliably navigate back/forth via browser history buttons. This is why I decided to use `js/history.replaceState` here as well ([b3100c6](https://github.com/nextjournal/clerk/pull/442/commits/b3100c6ee850b1d4a9985059b1fa9d49ca002c63)).

zampino:
Given the above comments and https://github.com/nextjournal/clerk/pull/437#issuecomment-1490021769 I think we can close this in favour of #437.

PR #448, opened 2023-04-03
Readme > usage: mention build!

holyjak:
I expected building of static html files to be included in the usage section (at least, since it is not mentioned in the Book).


mk:
@holyjak thank you! Will follow-up with mentioning the build functionality in the book as well.

PR #451, opened 2023-04-12
Internal Links

zampino:
Extend markdown syntax in prose to accept `[[internal-link]]` where `internal-link` works with: 

* a path to a notebook absolute or relative to project root (`[[notebooks/rule_30.clj]]`)
* a requirable namespace (`[[rule-30]]`)
* a qualified symbol pointing to a resolvable var (`[[rule-30/first-generation]]`)

We consider allowing this ambiguity in the syntax, and give precedence to paths pointing to existing files.


zampino:
Closing this as we decided to process all regular `[text](url)` links' hrefs from prose as per `clerk/doc-url` when sensible.

We might reserve the `[[wiki]]` syntax for more complex handling in the future, like linking by namespace/var or notebook title once we agree on a way to expose a canonical list of notebooks in the context of `clerk/show!`. @sritchie for instance, is suggesting to pass `clerk/serve!` the same options that are passed to `clerk/build!`.

PR #457, opened 2023-04-19
Window API & Taps Window

philippamarkovics:


mk:
I've removed the panel functionality in fdc294de6c99304af588d65be504920ff7e19fa6. We might want to revive this in the future but for now simplifying Clerk's render bundle is a higher priority for me.

PR #458, opened 2023-04-21
🏠 Home Page Improvements

joe-loco:
This improves the homepage including dark mode support. We also unify the index and make it usable in interactive mode so it's no longer needed to perform a `build!` to see it.

Add a header to the interactive mode that allows to navigate to both the index and home and allow to customize it.

* [x] Add index as a notebook
* [x] Fix multiple anchor click and history handlers after shadow reload
* [x] Add default index notebook to static build paths if none is given
* [x] Drop index view from static app 
* [x] Let `nextjournal.clerk.index` get paths from static build opts (currently always reads `deps.edn`
* [x] Rebuild header and (add footer ?) to in notebook-viewer @zampino 
* [x] Make header and footer customizable.
* [x] Only link to GitHub when sha is set (in static mode) @zampino 
* [x] Fix header in dark mode. @joe-loco
* [x] When there is no clerk alias with paths in `deps.edn` show error

Follow-up
* Build show file page. @joe-loco 
* Drop static_app.cljs in favour of render.cljs @zampino
* Unify homepage listing and static app index @joe-loco  
* follow up when a `deps.edn` guide lands in the book possibly in conjunction with a document section viewer

### DIFF
--- a/resources/stylesheets/viewer.css
+++ b/resources/stylesheets/viewer.css
@@ -137,7 +137,7 @@
 }
 @media (min-width: 960px){
     .notebook-viewer .code-viewer .cm-content {
-        @apply py-4 pl-12;
+        @apply pb-2 pl-12;
     }
     .notebook-viewer .code-listing {
         width: 48rem !important;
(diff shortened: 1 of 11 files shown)


zampino:
Tested on Clerk demo repo both over `:git/sha` and local jar, it works fine up to browser history navigation (URL is pushed correctly, but notebook is not shown) and that's broken against latest main as well. So we can merge and fix in a follow up?

Might be an issue with advanced compilation.

PR #470, opened 2023-05-03
Document-wide css class setting

zampino:
Allow to configure a `:nextjournal.clerk/doc-css-class` in namespace metadata to customise the main notebook viewer class. Takes string or vector values.

### DIFF
--- a/src/nextjournal/clerk/parser.cljc
+++ b/src/nextjournal/clerk/parser.cljc
@@ -48,13 +48,7 @@
 (defn visibility-marker? [form]
   (and (map? form) (contains? form :nextjournal.clerk/visibility)))
 
-(def doc-settings
-  "Settings allowed in the namespace's metadata map."
-  #{:nextjournal.clerk/doc-css-class})
-
 (def block-settings
-  "Settings allowed in form metadata or in viewer opttions. When specified in the namespace's meta map,
-  they apply to all forms in the notebook."
   #{:nextjournal.clerk/auto-expand-results?
     :nextjournal.clerk/budget
     :nextjournal.clerk/css-class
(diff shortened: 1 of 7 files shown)


src/nextjournal/clerk/parser.cljc
```diff
@@ -48,7 +48,13 @@
 (defn visibility-marker? [form]
   (and (map? form) (contains? form :nextjournal.clerk/visibility)))
 
+(def doc-settings
+  "Settings allowed in the namespace's metadata map."
+  #{:nextjournal.clerk/doc-css-class})
+
 (def block-settings
+  "Settings allowed in form metadata or in viewer opttions. When specified in the namespace's meta map,
```
mk, line 56: typo: `opttions`

src/nextjournal/clerk/parser.cljc
```diff
@@ -164,17 +170,20 @@
...
-                       false)
-   :block-settings (merge-with merge
-                               {:nextjournal.clerk/visibility {:code :show :result :show}}
-                               (parse-global-block-settings first-form))})
+  (merge
+   (when (ns? first-form) (select-keys (first (filter map? first-form)) doc-settings))
+   {:ns? (ns? first-form)
+    :error-on-missing-vars (parse-error-on-missing-vars first-form)
+    ;; TODO: unify with doc-settings, expose :n.c/toc-visibility user-facing
+    :toc-visibility (or (#{true :collapsed} (:nextjournal.clerk/toc
```
mk, line 178: We should either leave all doc keys namespaced or none so `:doc-css-class` is treated the same as `:toc-visibility`.

src/nextjournal/clerk/parser.cljc
```diff
@@ -48,7 +48,13 @@
 (defn visibility-marker? [form]
   (and (map? form) (contains? form :nextjournal.clerk/visibility)))
 
+(def doc-settings
+  "Settings allowed in the namespace's metadata map."
+  #{:nextjournal.clerk/doc-css-class})
```
mk, line 53: If we introduce this, I think `:error-on-missing-vars` and `:toc-visibility` are `doc-settings` as well?

mk (changes requested):
Would like to see these inconsistencies cleaned up and the build is failing.

src/nextjournal/clerk/parser.cljc
zampino: I'd be for either adding `:nextjournal.clerk/toc-visibility` to the list and exposing it user facing instead of `:n.c./toc` or  use `:n.c/toc` throughout the app and have it land to the render function as is and interpret it as _toc visibility_. Won't conflict with the unqualified `:toc`.

I'm also assusming we don't need the key-rename to `:nextjournal/X` for these keys.

src/nextjournal/clerk/parser.cljc
zampino: Yes, maybe. I first thought the names in the list are those whose values are parsed as-they-are and propagated to render fns, but that's wrong. 

I'd also assume the keywords in that list to be user-facing settings. So in that case we'd need to change the `:n.c/toc => :n.c/toc-visibility`.

PR #478, opened 2023-05-09
Propagate #inst with transform-fn mark-presented, fixes #476

### LINKED ISSUE #476: Serialization of java.time.Instant to viewer functions
During a ductile devcards POC I hit the case that when writing `#inst "2012"` and the underlying reader transforms that inst to a `java.time.Instant`, this wasn't serialized as a Date to the client where as with `#inst` producing a `java.util.Date` it did work.
I'm not sure if this should be solved inside of clerk, or as a user-land solution since it's a lossy conversion. But I guess translating `#inst "2012"` to `{:tag 'inst :value 2012}` is just as lossy since you don't know the type of the class anymore so you can't make any decisions based on that, so you might as well just convert to to a `js/Date` then?

A repro:

```
(clerk/with-viewer
  {:render-fn '#(pr-str %)
   :transform-fn clerk/mark-presented}
  {:a #inst"2023-10-08T00:00:00.000-00:00"})
```

borkdude:


borkdude:
@mk Should we add a test for this, if so what would be the best spot to do this?

mk:
@borkdude not really an easy way to test this, or at least no current setup. I think we'd need to test this comes across as the right type in a js env or do you see an easier way?

borkdude:
Thanks!

borkdude:
We can test this in notebook + mark-presented + an assert in the SCI viewer that will throw if the data wasn't a real data, probably.

PR #482, opened 2023-05-11
🤺 Amenable Fences

zampino:
The purpose of these changes is two-fold:

* Help solving annoying issues with the undesired evaluation of code in markdown fenced code blocks (currently always treated as clojure code meant for evaluation) like this repo's README.md. Also introduces a `skip` or `eval=false` (inspired by [RMarkdown/KnitR](https://www.rstudio.com/wp-content/uploads/2015/02/rmarkdown-cheatsheet.pdf)) annotaions to be used to signal a clojure fenced block should not be evaluated:
~~~
```clojure {:nextjournal.clerk/code-listing true}
'(1 2 "not evaluated" :but-still-highlighted)
```
~~~
* Prepare for highlighting inert markdown fenced code with suitable prosemirror styles. Don't highlight languages other than clojure for the moment.


mk:
I don't really like `skip` as I think it's ambiguous. `eval=false` is much clearer but then I also don't love that the default is true.

Looking at prior art, quarto seems to use `eval=false` under the execute key in all its markdown flavors:
* https://quarto.org/docs/reference/formats/markdown/gfm.html#execution
* https://quarto.org/docs/reference/formats/markdown/commonmark.html#execution
* https://quarto.org/docs/reference/formats/markdown/markua.html#execution

Ignoring this prior art (which I'm not sure we should do) I think I'd prefer something like `no-eval=true` or `code-listing=true`. In general I think it's good that we give some thought to how we want to support other clerk attributes to enable foldable code-listings or controlling width for example in languages that aren't clojure.

mk:
Leaning towards edn in the annotations so:

~~~markdown
```javascript {:visibility {:code :fold} :code-listing true}
41 + 1
```
~~~

Which we then namespace so `{:nextjournal.clerk/visibility {:code :fold} :nextjournal.clerk/code-listing true}`.

If we want to support a short-hand like for Clojure metadata we could read multiple times and merge, i.e.

~~~markdown
```javascript :code-listing {:visibility {:code :fold}}
41 + 1
```
~~~

PR #484, opened 2023-05-15
Fix check for extracting var from definition

### LINKED ISSUE #499: Return value from deflike macros is ignored
Reproduces with

```clj
(ns scratch.debug
  (:require [nextjournal.clerk :as clerk]))

(defmacro defmy [name body]
  `(do
     (def ~name ~body)
     (clerk/html [:h1.bg-green-200 "Defined: " [:span.font-mono.bg-slate-200 '~name]])))

(defmy stuff [1 2 3])
```

renders as 
CleanShot 2023-05-30 at 16 34 58@2x
instead of 
CleanShot 2023-05-30 at 16 34 06@2x

This is because the macro name starts with _def_ and it defines exactly one var.

zampino:
Fixes #499. At present, a user defined macro which starts with `def` and expands to code like

```clj
(defmacro defmy [name body]
  `(do 
     (def ~name ~body) 
     return-some-val))

(defmy stuff bla)
```

when called produces a cell result wrapped in a `{::clerk/var-from-def #'stuff}` while the intended return value is actually discarded.


src/nextjournal/clerk/analyzer.clj
```diff
@@ -85,7 +85,9 @@
 #_(rewrite-defcached '(nextjournal.clerk/defcached foo :bar))
 
 (defn deflike? [form]
-  (and (seq? form) (symbol? (first form)) (str/starts-with? (name (first form)) "def")))
+  (and (seq? form)
+       (symbol? (first form))
+       (contains? #{'def 'defonce 'defn} (first form))))
```
mk, line 90: The intent here was to actually account for user-defined macros that are like `def`s, so more than just those three concrete instances.

zampino: we could check the returned value is a var of the same name as the one inferred by the analyzer, to decide when to do the wrapping.

Issue #485, opened 2023-05-16
UI tests don't detect regression in cherry notebook


borkdude:
There is currently an error in the cherry notebook: 

https://snapshots.nextjournal.com/clerk/build/5e875e256a28a6deabf27bd6fc20f44cee5dad20/index.html#/notebooks/cherry.clj

but this doesn't get noticed by the UI tests.

This one seems to go unnoticed because there aren't any console.logs about the error.

borkdude:
Here is another example: https://snapshots.nextjournal.com/clerk/build/6d9501597a5bf8b2ae631421a8ea2f708e986602/index.html#/notebooks/cherry.clj

PR #486, opened 2023-05-16
Add global cherry evaluator setting + viewer opt

borkdude:
Add support for cherry as an alternative to sci to evaluate `:render-fn`s. You can change it per form (using form metadata or viewer opts) or doc-wide (using ns metadata) with `{:nextjournal.clerk/render-evaluator :cherry}`.


ui_tests/playwright_tests.cljs
```diff
@@ -64,15 +64,21 @@
                                          (not (str/ends-with?
                                                (.-url (.location msg)) "favicon.ico")))
                                 (swap! console-errors conj {:msg msg :notebook (.url page)}))))
+                     _ (.on page "pageerror"
```
zampino, line 67: Do these changes fix #485?

src/nextjournal/clerk/viewer.cljc
```diff
@@ -1349,26 +1352,29 @@
...
 
-(defn process-viewer [viewer]
+(defn process-viewer [viewer {:nextjournal/keys [render-evaluator]}]
   (if-not (map? viewer)
     viewer
-    (-> viewer
-        (dissoc :pred :transform-fn :update-viewers-fn)
-        (as-> viewer (assoc viewer :hash (hash-sha1 viewer)))
-        (process-render-fn))))
+    (let [evaluator (:render-evaluator viewer)]
```
zampino, line 1371: Maybe drop this let? you get a cleaner diff :-) I also checked `process-viewer` is called in just the one spot below, so this arity change should be fine.

ui_tests/playwright_tests.cljs
borkdude: It should be an improvement in general, but alas it didn't fix that error as the event wasn't triggered.  It can't hurt though.

See:
https://stackoverflow.com/questions/73448785/how-to-detect-uncaught-in-promise-typeerror-in-playwright/73459509#73459509

notebooks/cherry.clj
```diff
@@ -24,8 +27,17 @@
       [:pre
        (time (do (dotimes [_ 100000]
                    (js/Math.sin 100))
-                 (pr-str (interleave (cycle [1]) (frequencies [1 2 3 1 2 3])))))])
-   :evaluator :cherry}
+                 (pr-str (interleave (cycle [1]) (frequencies [1 2 3 1 2 3])))))])}
+  {:nextjournal.clerk/render-evaluator :cherry}
+  (+ 1 2 3 5))
+
+;; Since we set `:nextjournal.clerk/render-evaluator :cherry` on the ns meta, evaluation happens through cherry by default
```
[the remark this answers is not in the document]
borkdude: yes

Issue #489, opened 2023-05-17
Clerk Analyzer doesn't strip metadata from forms when hashing


craig-latacora:
The nextjournal.clerk.analyzer/hash-codeblock function will attempt to print out a form to generate a hash.  This will result in bad things happening if the form, when evaluated, tagged it's results with some metadata.  That metadata will now be applied to the form itself.  That's not much of a problem, except that metadata can be used to set the :tag of a value, and then a print-method can be defined for that tag.  That print-method body now will try and print the form (a sequence) and likely blow up.

Here is a quick patch that strips metadata before hashing the form.  This works, and I don't think that we need to be sensitive to changes in metadata.  However, perhaps the more correct mechanism would collect some part of the metadata and include that as input to the hash.

```clojure
(require '[nextjournal.clerk :as clerk])

;; This will compile and run just fine, but clerk errors when evaluating this as
;; a notebook
(defprotocol Foo
  (-foo [this]))

(defmethod print-method :foo [v ^java.io.Writer w]
  (.write w (str "FOO<" (-foo v) ">")))

;; When this is analyzed by clerk, you will get an exception:
;;
;; Unhandled java.lang.IllegalArgumentException
;; No implementation of method: :-foo of protocol: #'user/Foo found for class: clojure.lang.PersistentList

(defn gimme-a-foo []
  ^{:type :foo}
  (reify Foo
    (-foo [this]
      :baz)))

;; this should use the printer above
(gimme-a-foo)


;; evaluate this, and you will patch the
;; nextjournal.clerk.analyzer/hash-codeblock to strip metadata before printing
;; forms
;;
;; (in-ns 'nextjournal.clerk.analyzer)
;; (require 'clojure.walk)
;; (defn strip-form-meta [form]
;;     (clojure.walk/walk
;;      (fn [v]
;;        (if (or (instance? clojure.lang.IObj v)
;;                (instance? clojure.lang.IMeta v))
;;          (vary-meta v (constantly nil))
;;          v)) identity form))

;;   (defn hash-codeblock [->hash {:as codeblock :keys [hash form id deps vars]}]
;;     (when (and (seq deps) (not (ifn? ->hash)))
;;       (throw (ex-info "`->hash` must be `ifn?`" {:->hash ->hash :codeblock codeblock})))
;;     (let [hashed-deps (into #{} (map ->hash) deps)]
;;       (sha1-base58 (binding [*print-length* nil]
[...]

PR #490, opened 2023-05-17
Resilience in the face of metadata

### LINKED ISSUE #488: Analyzer sensitive to unknown class tags in forms
The analyzer will die on code that clojure runs just fine.  If I define a method which uses type tag metadata and the class is not known, it will compile and run fine, but clerk will error out when analyzing. 

Here is a minimal example that triggers it:

```clojure
(require '[nextjournal.clerk :as clerk])

;; this is a brute way to make the clerk analyzer resilient in the case of
;; an unknown class tag
;;
;; (alter-var-root #'nextjournal.clerk.analyzer/analyzer-passes-opts
;;                 (fn [v]
;;                   (assoc v
;;                          :validate/wrong-tag-handler (fn [tag node]
;;                                                        nil))))

;; This will compile and run just fine, but clerk errors when evaluating this as
;; a notebook.
(defmulti foo :tag)
(defmethod foo :baz [v] (pr ^Something v))
(foo {:tag :baz})
```

I think that ignoring those unresolvable tags is fine, but I am still learning clerk internals.

craig-latacora:
Closes #488 and #489 

Came across these while use clerk in a project that uses malli heavily.  I'm sure others will run into these as well.

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -12,6 +12,7 @@
             [clojure.tools.analyzer.ast :as ana-ast]
             [clojure.tools.analyzer.jvm :as ana-jvm]
             [clojure.tools.analyzer.utils :as ana-utils]
+            [clojure.walk :as walk]
             [multiformats.base.b58 :as b58]
             [multiformats.hash :as hash]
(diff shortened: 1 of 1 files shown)


craig-latacora:
This should be from a branch, not main.

PR #491, opened 2023-05-18
Metadata handling fixes in analyzer for #488 and #489

### LINKED ISSUE #488: Analyzer sensitive to unknown class tags in forms

craig-latacora:
This PR fixes the issues mentioned, by stripping metadata from forms before printing them to hash, and by not throwing an exception on an unrecognized class in a :type metadata tag.

closes #488 and #489

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -549,12 +549,21 @@
 #_(dep/immediate-dependencies (:graph (build-graph "src/nextjournal/clerk/analyzer.clj"))  #'nextjournal.clerk.analyzer/long-thing)
 #_(dep/transitive-dependencies (:graph (build-graph "src/nextjournal/clerk/analyzer.clj"))  #'nextjournal.clerk.analyzer/long-thing)
 
+(defn strip-form-meta [form]
+  (clojure.walk/postwalk
+   (fn [v]
+     (if (or (instance? clojure.lang.IObj v)
+             (instance? clojure.lang.IMeta v))
+       (vary-meta v (constantly nil))
+       v))
+   form))
+
 (defn hash-codeblock [->hash {:as codeblock :keys [hash form id deps vars]}]
   (when (and (seq deps) (not (ifn? ->hash)))
     (throw (ex-info "`->hash` must be `ifn?`" {:->hash ->hash :codeblock codeblock})))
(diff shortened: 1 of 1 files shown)


src/nextjournal/clerk/analyzer.clj
```diff
@@ -544,12 +549,21 @@
 #_(dep/immediate-dependencies (:graph (build-graph "src/nextjournal/clerk/analyzer.clj"))  #'nextjournal.clerk.analyzer/long-thing)
 #_(dep/transitive-dependencies (:graph (build-graph "src/nextjournal/clerk/analyzer.clj"))  #'nextjournal.clerk.analyzer/long-thing)
 
+(defn strip-form-meta [form]
+  (clojure.walk/postwalk
+   (fn [v]
+     (if (or (instance? clojure.lang.IObj v)
+             (instance? clojure.lang.IMeta v))
```
zampino, line 556: Hi @craig-latacora thank you for your pr!

If I can see correctly, a more correct check would be just `(instance? clojure.lang.IObj v)`. While it's unlikely that this `form` contains complex non-literal values, strictly looking at the interface definitions an object needs to implement `IObj` to be called `vary-meta` upon. For instance a var is IMeta but not IObj.

zampino: I've pushed your changes + some cleanup internally to https://github.com/nextjournal/clerk/compare/latacora/strip-meta?expand=1.

Issue #492, opened 2023-05-21
The instructions in the Book of Clerk for `:nextjournal.clerk/visibility` lead to errors when we run `clerk/show`


vedang:
The error is: `:nextjournal.clerk/visibility` expects a set of `#{:hide :show :fold ...}` but the documentation shows a map of `{:code :hide :results :show ... }`

PR #493, opened 2023-05-21
Update Clerk MVN version to the latest in the book

vedang:
The version of Clerk mentioned in the book right now is updated.

The problem this caused (for me) was with the
`:nextjournal.clerk/visibility` clause, which was a set in the previous version and is a map now. The documentation suggests using a map, but the execution fails since it expects a set.

Fixes: #492

### DIFF
--- a/book.clj
+++ b/book.clj
@@ -58,7 +58,7 @@
 ;; To use Clerk in your project, add the following dependency to your `deps.edn`:
 
 ;; ```edn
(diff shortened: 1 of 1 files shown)


mk:
Thanks for the PR and issue, good catch! I've followed up with 377dff9f49f2c942b155a35cc8c4c1e447a77931 so it should stay in sync from now on, thanks again!

PR #494, opened 2023-05-22
Fix clj-kondo warning with viewer-fn hook and fix static build

borkdude:


borkdude:
@mk I think this is ready for review/merge

resources/clj-kondo.exports/nextjournal/clerk/nextjournal/clerk/viewer.clj_kondo
```diff
@@ -2,11 +2,14 @@
...
-      (let [quoted-node {:tag :syntax-quote
-                         :children (:children quoted-node)}
-            fn-node (first (:children quoted-node))]
-        {:node fn-node}))))
+      (let [quoted-node
+            (with-meta
+              {:tag :syntax-quote
+               :children (:children quoted-node)}
+              (assoc (meta node)
+                     :clj-kondo/ignore [:unresolved-var]))]
```
mk, line 14: This is silencing the `:unresolved-var` warning now, correct? Would it also be possible to keep it, but always look it up via `cljs`?

borkdude: This isn't possible, but navigation in lsp still works since we implemented the .cljs fallback in December

mk: Ok, merged it. Would you be open to making this possible in hooks? Besides in Clerk we have a few other examples where this would be nice to have.

Issue #84, opened 2022-02-17
hashing inconsistent when form contains regexp


mk:
```clojure
(require '[nextjournal.clerk.hashing :as h])
(frequencies (into []
                   (comp (map (comp second vals h/hash h/build-graph h/parse-clojure-string)))
                   (repeat 200 "(fn [x] (clojure.string/split x #\"/\"))")))

;;=>  {"5drKysDWADNZJWwWfgWYs6TETrz7zx" 98, 
;;     "5dt92a8LBSfEScpUk3X7Y1L3vBDKJm" 2}
```

cc @behrica and thanks for bringing this up!

mk:
From https://clojure.org/guides/equality:
> Clojure regex’s, e.g. `#"a.*bc"`, are implemented using `Java java.util.regex.Pattern` objects, and Java’s equals on two Pattern objects returns `(identical? re1 re2)`, even though they are documented as immutable objects. Thus `(= #"abc" #"abc")` returns `false`, and `=` only returns `true` if two regex’s happen to be the same identical object in memory. Recommendation: Avoid using regex instances inside of Clojure data structures where you want to compare them to each other using `=`, and get true as the result even if the regex instances are not identical objects. If you feel the need to, consider converting them to strings first, e.g. `(str #"abc")` → `"abc"` (see [CLJ-1182](https://clojure.atlassian.net/browse/CLJ-1182))

PR #498, opened 2023-05-27
Set :regex true in analyzer/read-string

craig-latacora:
Without this, clojure.core.match macro expansion breaks.

If we're doing macroexpansion in the analyzer, we should make sure our read of the blocks matches the normal reader as closely as possible.

Minimal example
```
(ns notebook.matcher-fail
  (:require [clojure.core.match :as m]
            [nextjournal.clerk :as clerk]))

;; This form will compile fine, however, when analyzer and evaluating in clerk,
;; the macroexpandion of m/match will fail because that environment does not see
;; the method definition

(m/match ["foo" "bar"]
         ["foo" #"b.*"] "baz"
         :else "quux")
```

Note that it will break in the analyzer with or without the regex extension loaded.  That is not the issue, tho I did go down that garden path 8^)

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -105,7 +105,7 @@
                            :syntax-quote {:resolve-symbol tools.reader/resolve-symbol}
                            :readers *data-readers*
                            :read-cond :allow
-                           :regex #(list `re-pattern %)
+                           :regex true
                            :features #{:clj}
                            :auto-resolve (auto-resolves (or *ns* (find-ns 'user)))}))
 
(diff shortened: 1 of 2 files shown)


mk:
We need the form reading to be deterministic because we're using it as a key for anonymous top-level expressions, see #84 for the original issue.

@craig-latacora could you please create an issue outlining the problem so we can discuss alternative solutions there?

craig-latacora:
Will do, thanks for the context.  Still learning my way around the analyzer.  I had assumed that the hash-codeblock, which is based on hashing a pr-str of the form, was being used to track changes to forms.

PR #532, opened 2023-07-13
First cut of per-session state

mk:
This adds support for per-session state. When a document has a `:session` set, evaluation will happen in a separate synthetic namespace for the session. This paves the way to use Clerk for  multi-user web apps.


mk:
Might revisit this in the future.

PR #535, opened 2023-07-24
CSS compilation improvements

zampino:
* compile clerk stylesheet with tailwind against clojure source and notebooks on CI
* add it to asset map by default under the key `"/css/compiled-viewer.css"`
* use the compiled stylesheet as content during static build when opting in to compile css
* fix compiling CSS against custom local js bundles (see [discussion](https://nextjournal.slack.com/archives/C022KRVLAC9/p1677583840534479)) 
* bump tailwindcss to 3.3.3
* fix missing fonts under the `:exclude-js?` option


zampino:
I believe this approach won't work with a custom compiled js, unless

- we somehow detect from the asset map that we're building against a custom js bundle
- we open an extension point for adding source files containing tailwind classes

zampino:
> * we somehow detect from the asset map that we're building against a custom js bundle

we can compare clerk own asset map with the resource urls passed in the static app options: if they coincide we can prefer the source file.

zampino:
This is a bit odd. We need to restore `yarn.lock` or we run into a frontend hash change.

https://github.com/nextjournal/clerk/blob/b8f04651a2739fea5a4f3ee62e9205f40ff53c6d/.github/workflows/main.yml#L194-L200

We could be affected by https://github.com/npm/cli/issues/5126.

This should not be a problem for libraries consuming clerk.

zampino:
Tested against [clerk-demo](https://github.clerk.garden/nextjournal/clerk-demo/commit/ec8191f079d18f334affe5c01fa1e88993fa7209/) and https://github.com/nextjournal/markdown/pull/16.

zampino:
I leave a table of which Clerk commit on this branch is working/not working against clerk-demo:
- 3073470 ✅ [link](https://github.clerk.garden/nextjournal/clerk-demo/commit/e51c182d7127c6ec69cd4f8b3c86178dac8e02b9/) · Using a list of clj(c/s) code from resources as input 
- 54ca527 ✅  [link](https://github.clerk.garden/nextjournal/clerk-demo/commit/46cf52e94aba8a263a5f76e9396c61ed721ccb8a/notebooks/introduction/) · Using a .txt file from CAS produced with extractor as input
- f273e87 ❌ [link](https://github.clerk.garden/nextjournal/clerk-demo/commit/ad2debffb6e0289a2ad8fc1f9462ee54d8f42067/) · Using the compiled css from CAS as input.

What is a bit surprising is that f273e87 seems to produce the correct CSS in Clerk owns SSR+compiled-css notebooks (see build). While it's not working in clerk-demo (see link above / rule 30 notebook).

mk:
Let's wait on a cleaner tailwind extractor before we revisit this.

.github/workflows/main.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           target_url: https://snapshots.nextjournal.com/clerk/build/${{ github.sha }}
 
   ssr:
-    name: SSR smoke test build
```
zampino, line 167: nonsense

PR #541, opened 2023-08-03
Editor Improvements

zampino:
- [ ] add command bar
- [x]  allow to configure eval fns (notebook / string) e.g. for JVM side evaluation
- [x] buffer messages while web socket is not connected, send later when connected
- [ ] ~~extract eval + present-result in one function to allow caching and pagination on single expressions~~

Paired with [notebook](https://github.com/nextjournal/clerk-web-editor/blob/main/src/clerk_web_editor.clj), live on https://clerk-editor.live.clerk.garden.


zampino:
This has been mostly superseded by [folio](https://github.com/nextjournal/folio). Other stuff might be revived at some point in the future.

PR #561, opened 2023-10-11
Allow to disable welcome page

mk:
When passing the same `:paths` or `:paths-fn` arg as for `build!` to `serve!` Clerk will no longer the built-in welcome page and instead serve the generated or user-specified index page on root. This is a first step towards enabling easy sharing fully interactive Clerk documents including pagination &c.


src/nextjournal/clerk/webserver.clj
```diff
@@ -223,13 +240,25 @@
...
+
 (defn navigate! [{:as opts :keys [nav-path]}]
+  ;; TODO: perform `route-index` when needed, needs `:expanded-paths` for `v/route-index?`
   (show! opts (->file-or-ns (maybe-add-extension nav-path))))
 
 (defn prefetch-request? [req] (= "prefetch" (-> req :headers (get "purpose"))))
 
 (defn serve-notebook [{:as req :keys [uri]}]
-  (let [nav-path (subs uri 1)]
+  (let [opts (process-paths @!server)
```
zampino, line 259: Maybe call `process-paths` once within `serve!` below instead of each time here? `process-path` reads the filesystem if I can see correctly.

src/nextjournal/clerk.clj
```diff
@@ -50,7 +50,8 @@
                     :else
                     file-or-ns)
              doc (try (merge opts
-                             {:nav-path (webserver/->nav-path file-or-ns)}
+                             {:file file
```
zampino, line 53: Can we normalize `file` as string here? It seems to make troubles in the header: 

CleanShot 2023-10-12 at 16 33 56@2x

when calling `show!` against a namespace or a resource e.g.
```clojure
(clerk/show! (io/resource "nextjournal/clerk/tap.clj"))

(clerk/show! 'nextjournal.clerk.tap)
```

s `:file` consumed anywhere else?

deps.edn
```diff
@@ -73,6 +73,10 @@
                                                                  :exclusions [org.babashka/sci]}
                                io.github.nextjournal/clerk-slideshow {:git/sha "11a83fea564da04b9d17734f2031a4921d917893"}}}
 
+           :nextjournal/garden {:exec-fn nextjournal.clerk/serve!
+                                :exec-args {:index "book.clj"}
+                                :nextjournal.garden/aliases [:demo]}
+
```
zampino, line 79: I can [move this to its own deps.edn in a subdirectory](https://nextjournal.slack.com/archives/C047QGK2UG7/p1697119000296089) if we don't want it here.

src/nextjournal/clerk.clj
zampino: 🤔 ach nö, the doc returned from parse will always assoc a `:file` key with the same value. Maybe we should just handle it in the view? And there should no need to assoc it here?

zampino:
Also tested static builds locally, didn't see any regressions there.

src/nextjournal/clerk/webserver.clj
mk: Yes, calling this intentionally every time so changes to the filesystem are picked up without requiring a restart.

PR #575, opened 2023-11-21
Pass more Context to Viewer Functions + Cell Viewer

zampino:
* when a viewer's predicate is a map with a `:wrapped` key, its value is called against the whole wrapped value instead of just the value
* the contents of the cell are merged into the wrapped value on which viewer functions operate  
* a cell viewer is added to allow a.o. to control the visibility of result or code programmatically.

### DIFF
--- a/notebooks/viewers/context.clj
+++ b/notebooks/viewers/context.clj
@@ -1,9 +1,12 @@
 ;; # 🪬 Viewer Context
 
 (ns viewers.context
+  "Or how viewer API functions learn to take advantage of the document context"
+  {:nextjournal.clerk/visibility {:code :hide}}
   (:require [nextjournal.clerk :as clerk]
             [nextjournal.clerk.viewer :as viewer]))
 
+{::clerk/visibility {:code :show}}
 ;; A viewer's `:pred` function can perform viewer selection based on a value.
 
 ;; It would sometimes be useful to have more context available. Examples of this are:
@@ -27,10 +30,6 @@
 ;; metadata. Rejected this because it's invisible and doesn't work for
 ;; e.g. keywords.
 
-^::clerk/no-cache
-(clerk/add-viewers! [cljc-viewer])
-
-
 ;; Now what can we use this for? We can now, for example, create a
 ;; viewer in userspace that evalautes a given form in Clojure and sci,
 ;; allowing us to use it on both sides.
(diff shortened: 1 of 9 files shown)


notebooks/viewers/context.clj
```diff
@@ -0,0 +1,83 @@
...
+                          ;; default visibility check
+                          (let [{:keys [code? result?]} (viewer/->visibility cell)]
+                            (cond-> []
+                              code?
+                              (conj (viewer/cell->code-block-viewer cell))
+                              (or result?
+                                  (:ns? cell)
+                                  (-> cell :result :nextjournal/value (as-> n (when (number? n) (even? n)))))
+                              (conj (viewer/cell->result-viewer cell))))))))
+
```
zampino, line 69: @mk maybe an early feedback on this one 👆.

In some sense, controlling visibility is not "close to" the actual values which we want to display/hide. Unlike the usual customization of dedicated viewers matching different values, here we have just one spot in which to combine different predicates. Maybe good so anyway.

Function composability is not straightforward in this case: it would imply adding/removing to the returned vector. I tried to mitigate this via the helper functions `viewer/cell->code-block-viewer` and `viewer/cell->result-viewer` but one has to almost reimplement the original function. Any ideas on how to improve this further?

notebooks/viewers/context.clj
```diff
@@ -0,0 +1,78 @@
...
+(def cell-viewer
+  (update viewer/cell-viewer
+          :transform-fn comp
+          (clerk/update-val (fn [cell]
+                              (update-in cell [:settings ::clerk/visibility :result]
+                                         #(if (or (:ns? cell)
+                                                  (-> cell :result :nextjournal/value (as-> n (when (number? n) (even? n)))))
+                                            :show
+                                            %))))))
+
```
zampino, line 64: @mk maybe an early feedback on this one 👆.

In some sense, controlling visibility is not "close to" the actual values which we want to display/hide. Unlike the usual customization of dedicated viewers matching different values, here we have just one spot in which to combine different predicates. Maybe good so anyway.

zampino:
I checked that slideshow viewer still works, as this is the usual suspect when changing the shape of a presented notebook. Moreover #593 is not a regression from these changes, I'll fix that in a separate commit.

PR #592, opened 2023-12-15
Static Build: new `:package` option

zampino:
Introduce a new `:package` keyword for controlling `clerk/build!` artefacts. Deprecates `:bundle?` option. Clerk static apps now use client-side navigation by default for smoother page transitions.

### DIFF
--- a/src/nextjournal/clerk/render.cljs
+++ b/src/nextjournal/clerk/render.cljs
@@ -89,7 +89,7 @@
      status]]
    (when cell-progress
      [:div.w-full.bg-sky-100.dark:bg-purple-900.rounded.z-20 {:class "h-[2px] mt-[0.5px]"}
-      [:div.bg-sky-500.dark:bg-purple-400 {:class "h-[2px]" :style {:width (str (* cell-progress 100) "%")}}]])])5
+      [:div.bg-sky-500.dark:bg-purple-400 {:class "h-[2px]" :style {:width (str (* cell-progress 100) "%")}}]])])
 
 (defn connection-status [status]
   [:div.absolute.text-red-600.dark:text-white.text-xs.font-sans.ml-1.bg-white.dark:bg-red-800.rounded-full.shadow.z-30.font-bold.px-2.border.border-red-400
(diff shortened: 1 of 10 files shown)


src/nextjournal/clerk/render.cljs
```diff
@@ -745,25 +744,48 @@
...
-                       (throw (ex-info "Not Found" {:response r})))))
-            (.then (fn [edn]
-                     (set-state! {:doc (read-string edn)})
-                     (.pushState js/history #js {} ""
-                                 (cond-> path
-                                   (not (str/ends-with? path "/"))
-                                   (str "/"))))) ;; trailing slash is needed to make relative paths work
-            (.catch (fn [e] (js/console.error "Fetch failed" e ))))))))
+        (.. (fetch+set-state edn-path)
+            (then (fn [{:keys [ok]}]
```
mk, line 764: What's the `ok` here? Is it https://developer.mozilla.org/en-US/docs/Web/API/Response/ok? If yes I don't understand why we're accessing it through `:keys`. Should maybe be https://developer.mozilla.org/en-US/docs/Web/API/Response/text instead?

src/nextjournal/clerk/render.cljs
```diff
@@ -745,25 +744,48 @@
 (defn handle-initial-load [^js _e]
   (history-push-state {:path (subs js/location.pathname 1) :replace? true}))
 
-(defn click->xhr-request [e]
-  (when-some [url (some-> e .-target closest-anchor-parent .-href ->URL)]
+(defn fetch+set-state [edn-path]
+  (js/console.log "fetch" edn-path)
```
mk, line 748: ✂️

mk (approved):
Looks great, just one tiny comment.

src/nextjournal/clerk/render.cljs
zampino: I'm already using `.text` and `.-ok` from the fetch response object, but since I catch here, I need to signal failures or success to the consuming `.then` chains. See https://github.com/nextjournal/clerk/blob/69ed81bc8faa68fb49019973e3c09b71b30c09d2/src/nextjournal/clerk/render.cljs#L747-L755

PR #595, opened 2023-12-21
Fix vars from def in examples and comments

zampino:
🚨 just noticed this probably break fragments (check the tap inspector).


zampino:
Fixed with #598.

PR #602, opened 2024-01-05
Fix Editor

### LINKED ISSUE #601: Clerk Editor broke
A regression introduced with 413e3e5f4c790d08dd2d374076be80a535491596 (see [snapshot](https://snapshots.nextjournal.com/clerk/build/413e3e5f4c790d08dd2d374076be80a535491596/index.html#/notebooks/editor)).

zampino:
Fix #601.

### DIFF
--- a/src/nextjournal/clerk/render/editor.cljs
+++ b/src/nextjournal/clerk/render/editor.cljs
@@ -21,7 +21,7 @@
             [shadow.esm]))
 
 (defn eval-string
-  ([source] (sci/eval-string* (sci.ctx-store/get-ctx) source))
+  ([source] (eval-string (sci.ctx-store/get-ctx) source))
   ([ctx source]
    (when-some [code (not-empty (str/trim source))]
      (try {:result (sci/eval-string* ctx code)}
--- a/src/nextjournal/clerk/render/editor.cljs
+++ b/src/nextjournal/clerk/render/editor.cljs
@@ -24,7 +24,7 @@
   ([source] (eval-string (sci.ctx-store/get-ctx) source))
   ([ctx source]
    (when-some [code (not-empty (str/trim source))]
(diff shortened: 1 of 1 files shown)


src/nextjournal/clerk/render/editor.cljs
```diff
@@ -21,7 +21,7 @@
             [shadow.esm]))
 
 (defn eval-string
-  ([source] (sci/eval-string* (sci.ctx-store/get-ctx) source))
+  ([source] (eval-string (sci.ctx-store/get-ctx) source))
   ([ctx source]
    (when-some [code (not-empty (str/trim source))]
      (try {:result (sci/eval-string* ctx code)}
```
zampino, line 27: @borkdude not sure what's going on here. This first change wraps the result in a `:result` key which is needed by the view, but still doesn't fix the issue. Can you see why ns aliases from evaluating ns forms don't stick in the context?

src/nextjournal/clerk/render/editor.cljs
```diff
@@ -21,10 +21,10 @@
             [shadow.esm]))
 
 (defn eval-string
-  ([source] (sci/eval-string* (sci.ctx-store/get-ctx) source))
+  ([source] (eval-string (sci.ctx-store/get-ctx) source))
   ([ctx source]
    (when-some [code (not-empty (str/trim source))]
-     (try {:result (sci/eval-string* ctx code)}
+     (try (sci/eval-string* ctx code)
```
zampino, line 27: @borkdude not sure what's going on here. This change just adds the try/catch for arity-1, but the issue persists. Can you see why ns aliases from evaluating ns forms don't stick in the context?

zampino:
Fixed by @borkdude with 182864d.

PR #608, opened 2024-01-09
Don't eval forward-declared vars ahead of time

### LINKED ISSUE #454: `declare` should not show values
Screen Shot 2023-04-18 at 21 11 04

zampino:
Fix #454.

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -246,7 +246,7 @@
   (if var var id))
 
 (defn ->ana-keys [{:as analyzed :keys [form vars id]}]
-  (if (seq vars) vars [(->key analyzed)]))
+  (concat vars [(->key analyzed)]))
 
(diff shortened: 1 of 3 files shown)


src/nextjournal/clerk/analyzer.clj
```diff
@@ -246,7 +246,7 @@
   (if var var id))
 
 (defn ->ana-keys [{:as analyzed :keys [form vars id]}]
-  (if (seq vars) vars [(->key analyzed)]))
+  (conj (set vars) (->key analyzed)))
```
zampino, line 250: An expression like `(declare x y z)` has `vars` (and no `var`), in the previous code this implied `->analysis-info` map to never contain its analyze info as this would be overridden by the subsequent defs in

https://github.com/nextjournal/clerk/blob/75bd0f87b7c1da5a7d13677e9024fc30ef6a2fa1/src/nextjournal/clerk/analyzer.clj#L340-L343

This change should be safe, not sure yet if it's the optimal one.

zampino:
Covered by #629.

Issue #586, opened 2023-12-05
A Notebook throwing Exceptions leads to a blank Screen 


zampino:
This is not always reproducible, but clerk/show!ing a notebook like

```clojure
(ns repro 
  (:require [nexjtournal.clerk :as clerk]))

(clerk/html "x")

(/ 1 0)
```

causes the page to go blank with web console error

```
template.cljs:280 Uncaught Error: Assert failed: Invalid Hiccup form: [nil nil {:!expanded-at #object[reagent.ratom.RAtom {:val {:toc-open? true, :toc {}}}], :viewer nil, :path nil}]
 (in nextjournal$clerk$render$inspect_presented)
(valid-tag? tag)
```

PR #617, opened 2024-01-18
Ensure ex-data is always presentable

### LINKED ISSUE #616: Exceptions are not shown when watcher uses filter-fn
When the watcher is started with a filter function e.g.

```clojure
(clerk/serve! {:watch-paths ["notebooks" "src"] 
               :show-filter-fn #(clojure.string/starts-with? % "notebooks")})
```

when a notebook file is changed and contains e.g syntax or runtime errors then no exception is reported in the browser and the progress bar is stuck in analyzing state.

### LINKED ISSUE #622: Viewer crashes trying to present exceptions with function values in their data
Reproduces with

```clojure  
;; # 🐞
(ns scratch.debug
  (:require [nextjournal.clerk :as clerk]))

(ex-info "Boink 💥"
         {:foo 123 :boom (fn boom [x] x)}
         (RuntimeException. "no way"))
```
and web console reports known error: `Invalid symbol: scratch.debug/eval112468/boom--112469`.

zampino:
Ensure exceptions with ex-data are always inspectable.

Fixes #622. Fixes #616. Better fix for #586 than #587.

Screenshot 2024-01-22 at 18 12 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ Changes can be:
 
 * 🐜 Fix blank screen caused by react unmounting when an exception occurs during `clerk/show!`, fixes [#586](https://github.com/nextjournal/clerk/issues/586) @elken
 
+* 🐜 Fix crash using file watcher with `:show-filter-fn` in notebooks with exceptions, fixes [#616](https://github.com/nextjournal/clerk/issues/616).
+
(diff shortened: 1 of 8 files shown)


zampino:
Since it's not wrapped in a result viewer, the notebook exception view can't resolve elisions in the ex-data now (before, we'd just print all the data with the `inspect-wrapped-values` trick. I'll see if I can resolve that easily
Screenshot 2024-01-22 at 18 24 38

zampino:
Found another case that needs to be handled, ex-infos with data and without a cause.

PR #618, opened 2024-01-18
Read config values from ENV, Java properties and CLI args

Sohalt:
Some work towards unifying where configuration is read from.

Specifically this allows specifying the clerk cache location on the command line or with `exec-args`.

### DIFF
--- a/src/nextjournal/clerk.clj
+++ b/src/nextjournal/clerk.clj
@@ -561,14 +561,14 @@
   * the form of an anonymous expression"
   ([]
    (swap! webserver/!doc dissoc :blob->result)
-   (if (fs/exists? config/cache-dir)
-     (do (fs/delete-tree config/cache-dir)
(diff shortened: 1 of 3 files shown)


mk:
You can specify Java props from the Clojure cli as well so I don’t understand what problem this solves. Can we start with that?

Sohalt:
You can specify them on the clojure cli, when invoking it yourself, but not eg in `:exec-args` for use in garden or when calling clerk "as a library" from other clojure code.

mk:
An alias can specify jvm opts which should cover the garden use case? When using clerk as a library you can set the JVM prop programmatically.

jackrusher:
@mk The requirement to deal with both JVM opts and potential command line args come from me. Let's chat about it tomorrow.

Sohalt:
I'm just writing to properties for now.

PR #629, opened 2024-01-31
Analyzer Improvements

### LINKED ISSUE #66: dependency resolution in case of double defined defs is non-deterministic

### LINKED ISSUE #454: `declare` should not show values

### LINKED ISSUE #628: An edge case where the analyzer crashes
```clj 
(ns scratch.repro
  {:nextjournal.clerk/no-cache true})

(do
  (declare b)
  (def a 1))

(inc a)
```

Throws
```clj
Execution error (ExceptionInfo) at nextjournal.clerk.analyzer/hash-codeblock (analyzer.clj:565).
`->hash` must be `ifn?`
```
from 

https://github.com/nextjournal/clerk/blob/65988e062c8e9dc5e48111d05fec5350cf80e4c9/src/nextjournal/clerk/analyzer.clj#L562-L568

I find this error pretty confusing, as we should rather check that _all_ dependencies have been hashed at this point, and not that `->hash` is just a map. This could allow that _certain_ deps are not hashed and that the `nil` value lands in the `hashed-deps` set above.

The error above reproduces only if the node `a` is at the beginning of the list (hence precedes `declare`) in the topological sorting of the analyzer graph nodes, this is the not all the time the case, apparently.

This issue is due to the fact that we store the form 
```clj
(do
  (declare b)
  (def a 1))
```
under the key `a` in the analysis info map, while it probably should be stored under the block id only.

### LINKED ISSUE #634: Inline redefs crash during eval
```clojure
(ns scratch.repro)

(def a 1)

(def a (inc a))
```

crashes with `class clojure.lang.Var$Unbound cannot be cast to class java.lang.Number`. The first def is not evaluated at all since the second definition overrides the first in the analysis-info map at key `scratch.repro/a`.

zampino:
* Simplify Graph nodes (each code block appears in the graph just once by its block-id)
* Simplify hash of deps (only graph dependencies are used to compute the hash of a block)
* Improve assertion on missing hash for dependencies
* Take var snapshots during eval 
* Don't make expressions depend on forward declarations, but only on actual definitions
* Handle var redefinitions (all blocks with redefined vars are not cached)

Fixes #454 
Fixes #628
Fixes #66
Fixes #634

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -386,8 +386,6 @@
                              ;; `->analysis-info` :: { BlockId => Map }
                              ;; `var->block-id`   :: { Sym => Set<BlockId> }
                              (cond-> (-> state
-                                         (dissoc :doc?)
-
                                          (assoc-in [:->analysis-info block-id] analyzed)
                                          (track-var->block+redefs analyzed)
                                          (update :blocks conj (-> block
@@ -406,6 +404,8 @@
                                   :redefs #{}
                                   :blocks []))
                        (:blocks doc))
+
+         true (dissoc :doc?)
          doc? (-> parser/add-block-settings
                   parser/add-open-graph-metadata
                   parser/filter-code-blocks-without-form))))))
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -230,11 +230,14 @@
 (defn- circular-dependency-error? [e]
   (-> e ex-data :reason #{::dep/circular-dependency}))
 
-(defn- analyze-circular-dependency [state vars form dep {:keys [node dependency]}]
+(defn- analyze-circular-dependency [state vars form dep {:as _data :keys [node dependency]}]
   (let [rec-form (concat '(do) [form (get-in state [:->analysis-info dependency :form])])
         rec-var (symbol (str/join "+" (sort (conj vars dep))))
         var (first vars)] ;; TODO: reduce below
+    (when (not= dep dependency) (println :dep-mismatch dep dependency))
+    (when (not= var node) (println :node-mismatch var node))
     (-> state
+        (update :fixed-circular-deps assoc var rec-var dep rec-var)
         (update :graph #(-> %
                             (dep/remove-edge dependency node)
                             (dep/depend var rec-var)
@@ -243,25 +246,40 @@
 
 
 (defn ->key [{:as codeblock :keys [var id]}]
-  (if var var id))
+  ;; TODO: remove calls
+  #_(if var var id)
+  id)
 
 (defn ->ana-keys [{:as analyzed :keys [form vars id]}]
-  (if (seq vars) vars [(->key analyzed)]))
+  ;; TODO: remove calls
+  #_(if (seq vars) vars [(->key analyzed)])
+  [(->key analyzed)])
 
 #_(->> (nextjournal.clerk.eval/eval-string "(rand-int 100) (rand-int 100) (rand-int 100)") :blocks (mapv #(-> % :result :nextjournal/value)))
 
-(defn- analyze-deps [{:as analyzed :keys [form vars]} state dep]
-  (try (reduce (fn [state _var] ;; TODO: check if `_var` needs to be used
-                 (update state :graph #(dep/depend % (->key analyzed) dep)))
-               state
-               (->ana-keys analyzed))
-       (catch Exception e
-         (if (circular-dependency-error? e)
-           (analyze-circular-dependency state vars form dep (ex-data e))
-           (throw e)))))
+(defn dep->keys
+  "Inverse key-lookup from deps to key in analysis-info"
+  [{:keys [->analysis-info]} dep]
+  (or (not-empty
+       (keep (fn [[key {:keys [var vars]}]]
+               (when (contains? (cond-> (set vars) var (conj var)) dep)
+                 key)) ->analysis-info))
+      (when-not (deref? dep)
+        ;; TODO: check it is ok to remove deref deps
+        (list dep))))
 
+(defn- analyze-deps [{:as analyzed :keys [form vars]} state dep]
+  (try (reduce (fn [state k]
+                 (update state :graph dep/depend (->key analyzed) k))
+               state
+               (dep->keys state dep))
+       (catch Exception e
+      (if (circular-dependency-error? e)
+        (analyze-circular-dependency state vars form dep (ex-data e))
+        (throw e)))))
 
 (defn make-deps-inherit-no-cache [state {:as analyzed :keys [no-cache? vars deps ns-effect?]}]
+  ;; TODO: check wrt new keys
   (if no-cache?
     state
     (let [no-cache-deps? (boolean (some (fn [dep] (get-in state [:->analysis-info dep :no-cache?])) deps))]
@@ -287,7 +305,7 @@
   [sym]
   (str/includes? (name sym) ".proxy$"))
 
-(defn throw-if-dep-is-missing [{:keys [blocks ns error-on-missing-vars ->analysis-info file]}]
+(defn throw-if-dep-is-missing [{:as doc :keys [blocks ns error-on-missing-vars ->analysis-info file]}]
   (when (= :on error-on-missing-vars)
     (let [block-ids (into #{} (keep :id) blocks)
           ;; only take current blocks into account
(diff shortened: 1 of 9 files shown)


zampino:
Some edge cases still need to be addressed: 

```clojure 

(def a 1)

(def a (inc a))

a
```

the `analyzer/analyze`  fails to return `a` as dependency of the `(def a (inc a))`. This is because we set-subtract all introduced vars from the deps: 
https://github.com/nextjournal/clerk/blob/38a4653492bd47799f6dc5e6864e2b7eaf47a410/src/nextjournal/clerk/analyzer.clj#L187
Not doing that leads to other issues (e.g for multi-arity defns `(defn f ([x] (f {} x)) ([x opts] ...)`).


So when the first definition changes, this won't propagate to the value of `a` at the bottom. It's still an improvement wrt main, where the code above just crashes (#634).

zampino:
Second case: 

```clojure
(def a 1)

(inc a)

(def a :boom)
```

Similarly, this also crashes on main (#634) but evals correctly a first time on this branch. 

Subsequent `clerk/show!` fail when the consumer of the var `a` changes, since `a` is bound to a keyword, but `(def a 1)` didn't change. A similar behaviour to the "deref deps" might be applied here, but we need to decide when.

zampino:
Of course I didn't thought of the example:

```clojure 
(def a 1)
(when false (def a 2))
(inc a)
```

so also the current handling of redefinitions in the graph won't work reliably, I'm tempted to not cache definitions with the same name at all and let people decide (throw on redefs, warn+not-cache, not-cache).

zampino:
In this [other branch](https://github.com/nextjournal/clerk/compare/analyzer-improvements...analyzer-improvements--redefs-no-cache?expand=1) I've disabled caching for redefined vars (and therefore for their dependants). We might mark in the UI the blocks that are not cached and in case of the redefinitions explain why.

zampino:
Trying out passing info to code cells

https://github.com/nextjournal/clerk/assets/1078464/7f11c5db-3243-4681-bed5-1729a94e759a

src/nextjournal/clerk/eval.clj
```diff
@@ -163,12 +164,10 @@
     viewers
     (update :nextjournal/viewers eval)))
 
-(defn read+eval-cached [{:as _doc :keys [blob->result ->analysis-info ->hash]} codeblock]
-  (let [{:keys [form vars var]} codeblock
-        {:as form-info :keys [ns-effect? no-cache? freezable?]} (->analysis-info (if (seq vars) (first vars) (analyzer/->key codeblock)))
```
zampino, line 168: There was an ambiguity around under which key a form is stored inside the map `->analysis-info`. This used to cause e.g. #454. The caller of this function now replaces `codeblock` with `(get ->analysis-info (:id codeblock))`

src/nextjournal/clerk/analyzer.clj
```diff
@@ -188,8 +200,15 @@
       (seq deps) (assoc :deps deps)
       (seq deref-deps) (assoc :deref-deps deref-deps)
       (seq vars) (assoc :vars vars)
+      (seq vars-) (assoc :vars- vars-)
+      (seq declared) (assoc :declared declared)
```
zampino, line 204: Not using `:declared` vars at the moment, might be removed from here.

src/nextjournal/clerk/analyzer.clj
```diff
@@ -153,12 +168,8 @@
...
         analyzed (analyze-form {#'ana/macroexpand-1 mexpander} (rewrite-defcached form))
         nodes (ana-ast/nodes analyzed)
-        vars (into #{}
-                   (comp (filter (comp #{:def} :op))
-                         (keep :var)
-                         (map symbol))
-                   nodes)
-
+        {:keys [vars declared]} (get-vars+forward-declarations nodes)
+        vars- (set/difference vars declared)
```
zampino, line 172: we don't consider forward declarations among vars for building graph dependencies

src/nextjournal/clerk/analyzer.clj
```diff
@@ -234,39 +253,41 @@
...
            (throw e)))))
 
-
-(defn make-deps-inherit-no-cache [state {:as analyzed :keys [no-cache? vars deps ns-effect?]}]
-  (if no-cache?
-    state
-    (let [no-cache-deps? (boolean (some (fn [dep] (get-in state [:->analysis-info dep :no-cache?])) deps))]
-      (reduce (fn [state k]
-                (assoc-in state [:->analysis-info k :no-cache?] no-cache-deps?)) state (->ana-keys analyzed)))))
+(defn make-deps-inherit-no-cache [{:as doc :keys [blocks graph ->analysis-info]}]
```
zampino, line 279: this is now called once, after the whole graph has been built, to also consider edges across files (#633).

src/nextjournal/clerk/analyzer.clj
```diff
@@ -559,22 +602,24 @@
                            (vary-meta dissoc :type)))
                  form))
 
-(defn hash-codeblock [->hash {:as codeblock :keys [hash form id deps vars]}]
-  (when (and (seq deps) (not (ifn? ->hash)))
-    (throw (ex-info "`->hash` must be `ifn?`" {:->hash ->hash :codeblock codeblock})))
-  (let [hashed-deps (into #{} (map ->hash) deps)]
```
zampino, line 565: Hash is now computed using from graph nodes, rather than the block deps. I removed the assertion as it seemed confusing to me. The check should rather be that all deps has been hashed at this point.

zampino: > The check should rather be that all deps has been hashed at this point

If I introduce the same warning on `main`, by analyzing for instance `n.c.viewer` we get: 
```
Missing hash for dep clojure.core/class of form (defrecord ViewerEval [form])
Missing hash for dep clojure.core/declare of form (declare present+reset!)
Missing hash for dep clojure.core/class of form (defrecord ViewerEval [form])
Missing hash for dep nextjournal.clerk.viewer/get-viewers of form (defn ensure-wrapped-with-viewers ([x] (ensure-wrapped-with-viewers (get-viewers (get-*ns*)) x)) ([viewers x] (-> x ensure-wrapped (update :nextjournal/viewers (fn [x-viewers] (or x-viewers viewers))))))
Missing hash for dep clojure.core/class of form (defrecord ViewerFn [form])
Missing hash for dep nextjournal.clerk.viewer/add-viewers of form (defn apply-viewers* [wrapped-value] (when (empty? (->viewers wrapped-value)) (throw (ex-info "cannot apply empty viewers" {:wrapped-value wrapped-value}))) (let [viewers (->viewers wrapped-value) {:as viewer, viewers-to-add :add-viewers, :keys [render-fn transform-fn]} (viewer-for viewers wrapped-value) transformed-value (cond-> (ensure-wrapped-with-viewers viewers (cond-> (-> wrapped-value (dissoc :nextjournal/viewer) (assoc :nextjournal/applied-viewer viewer)) transform-fn transform-fn)) viewers-to-add (update :nextjournal/viewers add-viewers viewers-to-add)) wrapped-value' (cond-> transformed-value (-> transformed-value ->value wrapped-value?) (merge (->value transformed-value)))] (if (and transform-fn (not render-fn)) (recur wrapped-value') (-> wrapped-value' (assoc :nextjournal/viewer viewer) (merge (->opts wrapped-value))))))
Missing hash for dep nextjournal.clerk.viewer/present+paginate-children of form (defn present* [{:as wrapped-value, :keys [path !budget store!-wrapped-value], :nextjournal/keys [viewers]}] (when (empty? viewers) (throw (ex-info "cannot present* with empty viewers" {:wrapped-value wrapped-value}))) (when store!-wrapped-value (store!-wrapped-value wrapped-value)) (let [{:as wrapped-value-applied, :nextjournal/keys [presented?]} (apply-viewers* wrapped-value) xs (->value wrapped-value-applied)] (when (and !budget (not presented?)) (swap! !budget (fn* [%1] (max (dec %1) 0)))) (-> (merge (->opts wrapped-value-applied) (when (empty? path) (select-keys wrapped-value [:present-elision-fn])) (with-viewer (->viewer wrapped-value-applied) (cond presented? wrapped-value-applied (string? xs) (present+paginate-string wrapped-value-applied) (and xs (seqable? xs)) (present+paginate-children wrapped-value-applied) :else xs))) process-wrapped-value)))
Missing hash for dep nextjournal.clerk.viewer/transform-result of form (def result-viewer {:name (quote nextjournal.clerk.viewer/result-viewer), :render-fn (quote nextjournal.clerk.render/render-result), :transform-fn transform-result})
```
which I don't get on this branch.

src/nextjournal/clerk/analyzer.clj
```diff
@@ -1,5 +1,4 @@
 (ns nextjournal.clerk.analyzer
-  {:nextjournal.clerk/no-cache true}
```
zampino, line 2: Now that no-cache inheritance is fixed (#633) we need to remove this, otherwise expressions using a bunch of built-in viewers like `clerk/table` would not be cached.

src/nextjournal/clerk/analyzer.clj
```diff
@@ -234,39 +253,46 @@
...
+(defn make-deps-inherit-no-cache [{:as doc :keys [ns ->analysis-info blocks]}]
+  (let [nsname (when ns (str (ns-name ns)))]
+    (reduce (fn no-cache-inheritance-rf [state {:as analyzed :keys [id no-cache?]}]
+              (if no-cache?
+                state
+                (assoc-in state [:->analysis-info id :no-cache?]
+                          (boolean (some (fn [k]
+                                           (and ns (symbol? k) (= nsname (namespace k))
+                                                (get-in state [:->analysis-info k :no-cache?])))
+                                         (deps->block-ids state analyzed))))))
```
zampino, line 291: after discussion on #633, we now _ensure_ that `:no-cache?` does not propagate across files.

PR #649, opened 2024-06-05
Add license to pom.xml

borkdude:


mk:
Cherry-picked and applied with 352115ce862dffb2e4da2628a5743c9d4f1801ff which worked https://clojars.org/io.github.nextjournal/clerk/versions/0.16.1016. Thank you!

PR #655, opened 2024-07-08
Fix typo "tranform-fn" to "transform-fn"

vezwork:
Fixes a typo in the following sentence:

> ⚙️ Transform
When writing your own viewer, the first extension point you should reach for is `:tranform-fn`.


mk:
@vezwork thanks!

PR #662, opened 2024-07-30
Support empty tables

### LINKED ISSUE #661: Support empty tables
`(clerk/table [])` fails, but should show an empty table instead.

Sohalt:
Shows "no data" for empty tables, instead of an error.

2024-07-30-183954_985x193_scrot

Alternatively just show nothing?

Closes #661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changes can be:
 
 ## Unreleased
 
+* 💫 Make `clerk/table` work with empty tables.
 ...
 
 ## 0.16.1016 (2024-06-05)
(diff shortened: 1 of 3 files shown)


mk:
What does it look like for a table with only a `:head` and empty `:rows`? I'd try something more descriptive like `this table has no rows` in the message.

Sohalt:
> What does it look like for a table with only a `:head` and empty `:rows`?

It would look the same as for a completely empty table. I changed it to still show the headers. I also changed the message.

It now looks like this:

2024-07-31-101225_907x493_scrot

Not entirely sure about the `:nextjournal/width` default. For tables it defaults to `:wide`, but with the empty table that looks a bit weird imo. I also think it would be nice to have the "this table has no rows" message centered in the table, but I couldn't get that to work :/

mk:
@Sohalt cool, thanks. Can you add a changelog entry and then we merge it?

PR #663, opened 2024-08-05
Viewers can provide CLJS code in .cljs files

borkdude:
- Viewers can provide CLJS code in .cljs files
- Those .cljs files can be used in conjunction with the SCI browser REPL
- Viewer .cljs files can require other .cljs files (libraries)
- User notebooks can require viewers that use their own internal CLJS dependencies, like any other viewer


borkdude:
TODO:

- [ ] As a POC of this new feature, we will rewrite clerk-table-stats to use this.

PR #664, opened 2024-08-14
CLJS graph is no longer global but inferred from :require-cljs viewer key

borkdude:
- [x] The CLJS graph is no longer global, but inferred from `:require-cljs` keys in viewers
- [x] The `require-cljs` API call is no longer needed
- [x] Support `:require-cljs true` and infer namespace from `:render-fn` 

Will follow up with another PR to support unqualified render fns, if desired. Not 100% convinced if we really want this, so let's do this PR first.

### DIFF
--- a/src/nextjournal/clerk/cljs_libs.clj
+++ b/src/nextjournal/clerk/cljs_libs.clj
@@ -5,6 +5,7 @@
    [clojure.string :as str]
    [clojure.tools.namespace.dependency :as tnsd]
    [clojure.tools.namespace.parse :as tnsp]
+   [clojure.walk :as w]
    [edamame.core :as e]
    [nextjournal.clerk.viewer :as v]))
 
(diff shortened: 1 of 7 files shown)


src/nextjournal/clerk/cljs_libs.clj
```diff
@@ -68,36 +71,28 @@
...
+  (let [g (new-graph)]
+    (w/postwalk (fn [v]
+                  (when-let [cljs-ns (some-> v :nextjournal/viewer :require-cljs)]
+                    (require-cljs* g cljs-ns))
+                  v)
+                doc)
+    (into (omap/ordered-map :effects (let [resources (keep ns->resource (all-ns g))]
+                                       (mapv (fn [resource]
+                                               (let [code-str (slurp resource)]
+                                                 (v/->ViewerEval `(load-string ~code-str))))
```
zampino, line 87: Do we need to handle the evaluator=cherry case? Before it was implicitly handled during presentation.

PR #666, opened 2024-08-15
add react-dom to sci_env.cljs

tonsky:
Is it okay to add stuff like this? I need it in clerk-table-stats


mk:
Yep, since it’s already part of the bundle it doesn’t hurt to expose it via the sci.

PR #667, opened 2024-08-16
Fix cljs compilation for apps depending on Clerk render stack

zampino:
So far sci.nrepl is only mentioned in the `:dev` but it's used in clerk frontend.


mk:
@zampino right, been meaning to make this change as well, thank you! I undid the config change which doesn't change behavior as I understand it (we can't merge false) so I don't think it adds clarity.

PR #669, opened 2024-08-21
Improve analyzer graph dependencies

### LINKED ISSUE #660: Analyzer crashes on specter's caching implementation (analyzing forms that create new defs)
I created a full reproducer here: https://github.com/latacora/clerk-analyzer-specter-bug

The bug seems to be caused because Specter has a macro expansion that itself new vars (for local cache purposes).

For search purposes, the error looks like this:

```
Unhandled clojure.lang.ExceptionInfo
Hash is missing on dependency 'whatever/pathcache1234' of the form '(defn f [x] (sr/select [...])) 'in foo.bar.baz
```

I wonder if the analyzer could just compute the hash when this happens?

As a workaround, use `select*`, which is the version with no fancy caching.

zampino:
Fixes #660

* don’t fail when a form depends on references created at runtime, e.g. when a macro interns vars?
* Don't throw when hash of a dependency is not found
* Add means to record missing hashes on dependencies
* Fix order dependence in analysis: create graph edges in one go, only after var->block-id is resolved for all namespaces (this fixes scenarios like https://github.com/nextjournal/clerk/issues/660#issuecomment-2298832010)

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -340,18 +340,11 @@
     identity))
 #_ (ns-resolver *ns*)
 
-(defn analyze-doc-deps [{:as doc :keys [redefs blocks ->analysis-info]}]
-  (reduce (fn [state {:as info :keys [id deps vars- no-cache?]}]
-            (reduce (partial analyze-deps info)
-                    (cond-> state
-                      ;; redefinitions are never cached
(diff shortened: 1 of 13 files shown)


src/nextjournal/clerk/analyzer.clj
```diff
@@ -340,18 +340,11 @@
     identity))
 #_ (ns-resolver *ns*)
```
zampino, line 342: what the hell

PR #670, opened 2024-08-23
Improve bounded count check

zampino:
An [ISeq is always an IPersistentCollection](https://github.com/clojure/clojure/blob/48b1fe5b50d48603f2c1fbd38223a7284520d1ed/src/jvm/clojure/lang/ISeq.java#L18) but the converse is not true. 

Vectors and maps are not ISeq and for instances of such, the current protocol implementation doesn't check the limit on the whole collection, but only on its items.


borkdude:
LGTM

PR #676, opened 2024-08-28
Let sci.nrepl reuse Clerk's websocket

mk:
Allowing us to leverage Clerk's reconnect handling etc. Goes along with https://github.com/babashka/sci.nrepl/pull/7.

### DIFF
--- a/notebooks/render_fns.cljs
+++ b/notebooks/render_fns.cljs
@@ -4,3 +4,7 @@
 
 (defn motion-div [props]
   [:> (.-div motion) props])
+
+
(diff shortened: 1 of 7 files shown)


notebooks/render_fns.cljs
```diff
@@ -4,3 +4,7 @@
 
 (defn motion-div [props]
   [:> (.-div motion) props])
+
+
+
+(js/console.log :foo!!)
```
borkdude, line 10: probably can get rid of this

PR #686, opened 2024-09-06
fix docstring inconsistency in build!

otfrom:
:single-page should be :single-file

:gnome:


mk:
@otfrom thank you!

otfrom:
Thanks for leaving a small one like this so that I can feel useful <3

Issue #531, opened 2023-07-07
`build-graph` tries to look into non-closure code


foredu:
`build-graph` uses location provided by symbol's meta information and assumes Clojure source code. In some cases it isn't. E.g. *libpython-clj* refers to Python source code. Watching a foreign source code location might be still useful for watching file changes.

This:

```
(ns non-clojure-source
  (:require [libpython-clj2.python :refer [initialize!]]
            [libpython-clj2.require :refer [require-python]]))

(initialize!)
(require-python 'numpy)
(numpy/loadtxt "/dev/null")
```

fails with `Invalid symbol: BagObj:`, which is related to a Python class definition in the source file ("numpy/lib/npyio.py").

PR #688, opened 2024-09-12
Ignore python files introduced by libpython-clj

philippkueng:
While working with Clojure, Clerk and OpenCV I noticed that while requiring a python library works as soon as one is invoking a function it'll report an error in Clerk.

Screenshot 2024-09-12 at 08 35 35

I've extended the file parsing function adding a special case for python files to ignore them, similar to how markdown files are treated specially.

Let me know if you'd like to see anything else to have this change merged.


philippkueng:
Forgot to mention, here's the reproduction of the issue: https://github.com/philippkueng/clerk-issue-531
And here's an issue from last year most likely pointing to the same issue: https://github.com/nextjournal/clerk/issues/531

philippkueng:
Hi maintainers of clerk, 

I've been wondering if you had a chance to look at the PR? If so, are you not happy with the proposed solution or the use-case it's trying solve for or something else? I'd be totally happy adapt.

mk:
Hi @philippkueng,

I've now gotten around to implementing turning off Clerk's analysis completely when caching is disabled in cc07b80fb252d0eba566e8cbaaa2ef55dd055781. For now my recommendation is to turn off Clerk's caching with `{:nextjournal.clerk/no-cache true}` on the ns metadata. Let me know if this works for you.

My plan is to follow-up with an alternative, simpler analyzer that should also work for libpython.

philippkueng:
@mk thank you Martin for you change and your help with this. It's very much appreciated and it completely fixes the problems I was experiencing. Many thanks

PR #701, opened 2024-10-07
Improve error ux by showing ex-cause of error when set

mk:
The ux for eval errors could be pretty bad in case there was rich viewers in the doc. This improves it.

Before:
Screenshot 2024-10-07 at 16 02 02

After:
Screenshot 2024-10-07 at 16 02 28


mk:
Superseded by 3835df3f1635ba0be8a2e68dd3b86bd196071da9.

PR #702, opened 2024-10-07
Check ns is required before sci-copy-nss

tonsky:


mk:
This will be fixed when we upgrade sci to include https://github.com/babashka/sci/commit/99b862893280f9829c2a365d0e323a571b00fc17.

PR #710, opened 2024-10-18
Fix sync atom with "big data"

borkdude:


borkdude:
Feel free to still add the notebook for testing purposes.

PR #711, opened 2024-10-19
Delay viewer eval

mk:
Currently doing eval on read, this leads to problems: 
    * no control over when eval happens but sync atoms need to be interned before eval
    * no editscript diff for sync atom values
    * can't show error when clerk-eval fails because errors need to be reset before read


mk:
Superseded by 126ddaf2b9338e0644e91a8345ebb23724fc1d26... 08611b2b295af33bd348b7ce3895ddab7a528b86.

PR #725, opened 2024-10-23
Fix #718: expose spec

### LINKED ISSUE #718: ClojureScript namespaces missing from required deps in .cljs files loaded via `:require-cljs`
Since the most recent version it is possible to load .cljs files from Clerk viewers using `:require-cljs true`. This works fine, but when I try to require a library in such a file (added in my deps.edn), several namespaces normally included in ClojureScript are reported missing: "Could not find source for CLJS namespace x"

When testing it with a library of my own[1], the following namespaces are missing:

- cljs.core
- clojure.math
- clojure.pprint
- clojure.spec.alpha
- clojure.spec.gen.alpha
- goog  (yeah, just "goog")
- goog.math.Long

In our discussion on Clojurians Slack yesterday, @mk  added `clojure.math`: https://github.com/nextjournal/clerk/commit/319c00221862ee3fb975a3b75bdc1d6b5d9e0754 which worked fine when I tested it. Maybe the other namespaces can also be added to increase compatibility?

Here is a test repo with my library as a dependency to reproduce the issue: https://github.com/formsandlines/clerk-cljs-test If all namespaces are included, the `test-viewer` in notebooks/index.clj using the `render-test` function from src/render.cljs should display `:I` without any errors.

---

[1]: https://clojars.org/eu.formsandlines/formform (works both in clj and cljs) - I would like to render some visualizations I created as web components that need functions from that lib, so I hope to get it working there

borkdude:


borkdude:
Note that spec was already required by js-interop here:

https://github.com/applied-science/js-interop/blob/ffcc9cbd9981b18acd45b2d017a57d1ef2278c73/src/main/applied_science/js_interop/destructure.cljc#L5

mk:
Since adding things is hard to take back I'm wondering what the impact of it is compared to not depending on it. Is it fair to say that using`js-interop` in sci is an antipattern because there's no compile-time macros which speed things up in js-interop?

borkdude:
> Is it fair to say that usingjs-interop in sci is an antipattern because there's no compile-time macros which speed things up?

Some people want it for convenience. It's also available in nbb for example.

PR #727, opened 2024-10-29
Fix rendering of numbers whose magnitude is greater than 2^53

krzsztf:
Only numbers with magnitude no greater than 2^53 are precisely representable in JavaScript Number type.


mk:
@krzsztf thank you!

PR #736, opened 2024-12-10
Notebook with SCI + js/import

borkdude:
- Bumps SCI such that interop on locals work (see notebook)
- Dedupes SCI version between builds (always use :sci alias)

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "w3c-keyname": "2.2.4"
   },
   "devDependencies": {
-    "process": "^0.11.10"
+    "process": "^0.11.10",
+    "shadow-cljs": "^2.28.20"
   }
 }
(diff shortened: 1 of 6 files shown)


borkdude:
@mk I went a head and merged this one. It fixes one issue with SCI but also one issue with the build: due to the order of aliases, it always used the SCI version specified in the demo alias. I removed the SCI version from :demo and modified the build to always use the `:sci` aliases such that we always have a predictable version of SCI (namely the one specified in `render/deps.edn`.

package.json
```diff
@@ -36,6 +36,7 @@
     "w3c-keyname": "2.2.4"
   },
   "devDependencies": {
-    "process": "^0.11.10"
+    "process": "^0.11.10",
+    "shadow-cljs": "^2.28.20"
```
mk, line 40: @borkdude is this what triggers all the additions in yarn.lock? Do we need this?

borkdude: Done!

PR #774, opened 2025-10-27
Exercise transitive deps in test

borkdude:


mk:
@borkdude thanks! I've updated the code to use our existing `analyze-doc-deps` fn in order to get the handling of circular dependencies.

PR #780, opened 2025-11-24
add option to disable inline formulas in markdown

### LINKED ISSUE #775: Dollar signs in comments incorrectly trigger TeX rendering
Hi! Clerk has been excellent overall, but I've encountered an issue with dollar signs in comments being interpreted as TeX delimiters.

**Problem:**

When writing comments that include dollar amounts, Clerk treats them as TeX expressions. For example:

```clojure
;; showing the dual-fuel advantage grows from ~$50K at low volatility to ~$115K at high volatility.
```

This renders using the TeX renderer instead of displaying as plain text.

Image

**Current Workaround:**

I need to escape each dollar sign:

```clojure
;; showing the dual-fuel advantage grows from ~\$50K at low volatility to ~\$115K at high volatility.
```

This becomes tedious quickly since we write dollar amounts frequently and never use TeX in our codebase.

**Question:**

Is there a way to disable TeX rendering in comments, or a configuration option to prevent dollar signs from triggering the TeX parser? Any suggestions for a better solution would be appreciated.

borkdude:
Allow setting `{:nextjournal.clerk/markdown {:disable-inline-formulas true}}` as ns metadata to disable markdown parsing of $-delimited inline formulas. Fixes #775.

### DIFF
--- a/src/nextjournal/clerk/parser.cljc
+++ b/src/nextjournal/clerk/parser.cljc
@@ -472,9 +472,8 @@
                                     :settings (merge-settings next-block-settings (parse-local-block-settings form))
                                     :text nstring
                                     :form (add-loc opts loc form)
-                                    :loc loc}
-                        ns? (ns? form)]
-                    (when ns?
+                                    :loc loc}]
+                    (when (ns? form)
                       (eval form))
                     (cond-> (-> state
                                 (assoc :add-comment-on-line? true)


mk:
> Do we need to support markdown config in other places, e.g. via a system property? Or one system property with a clerk config to rule them all?

I need to think more about this. Would leave this pr to only add the notebook metadata. We can start collecting use cases for a global clerk config in an issue and then think about how to solve it separately.

src/nextjournal/clerk/parser.cljc
```diff
@@ -471,16 +470,18 @@
...
                       (eval form))
                     (cond-> (-> state
                                 (assoc :add-comment-on-line? true)
                                 (update :nodes rest)
                                 (assoc :block-settings next-block-settings)
                                 (update :blocks conj (add-block-id code-block)))
                       (not (contains? state :ns))
-                      (assoc :ns *ns*)))
+                      (assoc :ns *ns*)
+                      ns? (update-in [:md-context :opts] merge (:nextjournal.clerk/markdown (meta *ns*)))))
```
mk: Think we want to also support setting this in a `nextjournal.clerk.parser/settings-marker?` which is a top-level map with one of the `nextjournal.clerk.parser/block-settings` keys set.

borkdude:
> We can start collecting use cases for a global clerk config in an issue and then think about how to solve it separately.

We probably don't even need an issue, we can wait for people to create one after they want more than just the notebook annotation.

Restricting to just notebook metadata for now sounds good.

PR #803, opened 2026-02-05
Improve presentation performance

### LINKED ISSUE #731: Presentation performance notes
Benchmarking using the Book of Clerk @tonsky and I found that:

* [ ] image handling is ~25%, should cache it with a type holding the memory image and encoded image
* [ ] `prepend-required-cljs` is ~10%, can probably record the viewers at presentation time
* [ ] bounded-count-limit is very high with 1 million, let's lower it

borkdude:
Improves presentation performance by
- avoiding postwalks for `prepend-required-cljs` and `process-blobs`
- caching the image processing in memory
- lowering the bounded count limit from 1.000.000 to 1.000

Closes #731.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Changes can be:
   
 * 🐞 Always analyze doc (but not deps) when no-cache is set, @borkdude, fixes [#781](https://github.com/nextjournal/clerk/issues/781).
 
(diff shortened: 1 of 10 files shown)


mk:
One thing I noticed when testing this is that we do still spend quite a bit of time in `read-image` during eval, especially when loading images over the network e.g. in `notebooks/viewers/image.clj`. This is out of scope for this but I wanted to mention it as a potential future optimization.

PR #810, opened 2026-04-20
Fix `build-graph` crash on non-Clojure source files (#531)

### LINKED ISSUE #531: `build-graph` tries to look into non-closure code

borkdude:
Skip `parse-file` for sources that aren't `.clj`/`.cljc`/`.cljs`/`.bb` (e.g. `.py` files from libpython-clj vars); hash them opaquely like jars.

Fixes #531.

### DIFF
--- a/src/nextjournal/clerk/analyzer.clj
+++ b/src/nextjournal/clerk/analyzer.clj
@@ -550,8 +550,11 @@
                                  (let [jar? (or (nil? source)
                                                 (str/ends-with? source ".jar"))
                                        gitlib-hash (and (not jar?)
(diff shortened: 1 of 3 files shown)


borkdude:
Lint issue is unrelated and fixed in #809

PR #814, opened 2026-04-25
Fix pagination

### LINKED ISSUE #813: Table expansion returns data from wrong block after #803
After #803 (Improve presentation performance, `8d89c6fc`), expanding a paginated `clerk/table` can return rows from a *different* table or def in the same notebook.

### Repro

A notebook with two tables (e.g. a `clerk/table` call followed by a `def` with `^{::clerk/viewer clerk/table}` metadata). When the first table is expanded past its initial page, the expanded rows come from the second table's data.

### Root cause

The `notebook-viewer` transform now merges `(->opts wrapped-value)` into the doc before calling `process-blocks`:

```clojure
;; notebook-viewer :transform-fn (new)
(update :nextjournal/value (fn [value]
                              (process-blocks viewers (merge (->opts wrapped-value) value))))
```

This means the doc now carries the notebook-level `store!-wrapped-value`, `present-elision-fn`, and `store!-viewer` functions. `process-blocks` extracts these via `(->opts doc)` and merges them into each block before calling `present`:

```clojure
(let [opts (->opts doc)]   ;; includes notebook-level store!-wrapped-value, present-elision-fn
  ...
  (map (comp #(present (merge opts %)) ...)))
```

Inside `present`, the merge order on lines 1845-1859 is:

```clojure
(merge {:store!-wrapped-value <fresh-per-block-fn>
        :present-elision-fn   <fresh-per-block-fn>
        ...}
       (make-!budget-opts opts)
       opts)   ;; <-- opts from the block, which now carries the notebook-level fns
```

The notebook-level `store!-wrapped-value` and `present-elision-fn` (from `opts`) are merged **last**, overwriting the fresh per-block versions. All blocks then share a single `!path->wrapped-value` atom, so paths like `[0]`, `[1]`, etc. collide across blocks. When the client fetches an elision, `present-elision*` looks up the path in this shared atom and can return data from a different block.

### Before #803

The old `notebook-viewer` was:
```clojure
(update :nextjournal/value (partial process-blocks viewers))
```

And the old `process-blocks` didn't extract or pass opts — each block's `present` call created its own `!path->wrapped-value` atom.

### Suggested fix

Exclude `store!-wrapped-value`, `present-elision-fn`, and `store!-viewer` from the opts that `process-blocks` passes into blocks. These should always be created fresh per `present` call. Alternatively, change the merge order in `present` so that the fresh per-block fns take precedence over inherited opts.

### Affected version

Introduced in `8d89c6fc` (#803). The parent `cfdcb9f3` does not have this bug.

mk:
Fixes #813.

### DIFF
--- a/src/nextjournal/clerk/viewer.cljc
+++ b/src/nextjournal/clerk/viewer.cljc
@@ -1361,7 +1361,7 @@
    :transform-fn (fn [{:as wrapped-value :nextjournal/keys [viewers]}]
                    (-> wrapped-value
                        (update :nextjournal/value (fn [value]
-                                                    (process-blocks viewers (merge (->opts wrapped-value) value))))
+                                                    (process-blocks viewers (merge (select-keys wrapped-value [:store!-viewer]) value))))
                        mark-presented))})
 
 (def render-eval-viewer
--- a/test/nextjournal/clerk/viewer_test.clj
+++ b/test/nextjournal/clerk/viewer_test.clj
@@ -110,12 +110,12 @@
                 (clerk/table {:row (mapv #(str \"t1-\" %) (range 30))})
                 (clerk/table {:row (mapv #(str \"t2-\" %) (range 30))})"
           table-results (->> (eval/eval-string code)
-               view/doc->viewer
-               :nextjournal/value
-               :blocks
-               (mapcat :nextjournal/value)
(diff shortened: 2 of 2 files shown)


borkdude:
@mk LGTM. Maybe we should add a test for this? Can make one tomorrow.

PR #817, opened 2026-07-19
Assert the front-end hash is not blank

borkdude:


borkdude:
@mk This test was always passing due to the function always returning a string which is truthy

PR #822, opened 2026-09-14
Normalize edamame opts

borkdude:
@mk Can you please merge this? edamame will start throwing in `parse-next` when options aren't normalized first. This is for perf and correctness reasons.


borkdude:
Oops, I accidentally clicked on "Request review" when Copilot was selected, my bad.

